PR #4, opened 2017-02-10
Fix kernel integration

bdunogier:
> [EZP-26982](https://jira.ez.no/browse/EZP-26982)

Fixes the integration of the package into ezpublish-kernel.

### Tasks
- [x] Make `AppCache` use `TagAwareStore`
- [x] Make sure that purge requests use the `key` header and not `xkey`
- [x] Write a specification for `KernelPass`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,13 +25,24 @@ Add the package to `app/AppKernel.php`, *before* the EzPublishCoreBundle declara
 The package will replace the services from the kernel, thus enabling the new features, such as multi-tagging.
 
 The application cache class needs to be customized. If you haven't changed the `AppCache` class, you can do so
-by setting the `SYMFONY_HTTP_CACHE_CLASS` environment variable:
+by setting the `SYMFONY_HTTP_CACHE_CLASS` environment variable for your PHP or web server user.
+If you use your own `AppCache` class, you will have to make it to extend from this class instead
+of from the CoreBundle's.
+
+For PHP's internal server:
 
     export SYMFONY_HTTP_CACHE_CLASS='EzSystems\PlatformHttpCacheBundle\AppCache'
 
-Do not forget to restart your web server.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -24,6 +24,13 @@ Add the package to `app/AppKernel.php`, *before* the EzPublishCoreBundle declara
 
 The package will replace the services from the kernel, thus enabling the new features, such as multi-tagging.
 
+The application cache class needs to be customized. You can do so by setting the `SYMFONY_HTTP_CACHE_CLASS`
+environment variable:
+
+    export SYMFONY_HTTP_CACHE_CLASS='EzSystems\PlatformHttpCacheBundle\AppCache'
+
+Do not forget to restart your web server.
```
andrerom: alternatively user needs to extend it, for use cases where they need to add own customizations.

bdunogier: How would you phrase that ? I expect that somebody who has extended it knows what needs to be done at this point.

glye: Keep it simple: "Alternatively, if you need to make your own customizations, you should extend this class and set `SYMFONY_HTTP_CACHE_CLASS` accordingly."

andrerom: "and set SYMFONY_HTTP_CACHE_CLASS accordingly."

if you extend class from this package you don't need to set the env variable, as you'll use the default AppCache class _(which will be the one to extend the class in this package)_.

src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -9,18 +9,14 @@
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-/**
- *
```
glye: Could have added some doc instead ;)

bdunogier: Docs added before merging :)

PR #5, opened 2017-02-10
RFC: Added behat scenario for view cache

bdunogier:
Aims at describing the behavior one expects from content views HTTP cache in eZ Platform.

## Feedback I'm looking for
- Actual target audience for these scenarii ("one" in the sentence above)
- Comments about the domain for this target audience being bad in sentences (too specific, not specific enough, etc)
- _Unknown_ missing scenarii :)

## Known missing scenarii
- [ ] siblings purge
- [ ] parent purge
- [ ] relations purge
- [ ] path purge

### DIFF
--- a/features/view_cache.feature
+++ b/features/view_cache.feature
@@ -1,0 +1,20 @@
+Feature: Views are cached by the reverse proxy, and purged by the repository.
+
+  Background:
+    Given that Symfony's reverse proxy is enabled
+      And View cache is enabled
+      And Purge type is set to local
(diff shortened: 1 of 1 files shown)


bdunogier:
Ping @andrerom, @yannickroger.

andrerom:
> Actual target audience for these scenarii ("one" in the sentence above)

Developers using the Repository API /  view layers perhaps? To cover the behaviour of slots somehow.

PR #11, opened 2017-03-20
Use correct cache manager service name

emodric:


bdunogier:
Oops...

I guess that it just didn't work before, and now it does ? In that case +1 :)

emodric:
I didn't try it, but it seems ilogical to use a deprecated service when there's a new implementation here, no?

PR #13, opened 2017-08-04
EZP-27713: As a Developer I want HttpCache to handle RemoveTranslationSignal

alongosz:
Implements [EZP-27713](https://jira.ez.no/browse/EZP-27713)
----

This PR implements HttpCache `RemoveTranslationSlot` for `RemoveTranslationSignal` introduced in ezsystems/ezpublish-kernel#2031.

Similar slot for deprecated EzPublishCoreBundle HttpCache was introduced in ezsystems/ezpublish-kernel#2062.

**TODO**:
- [x] Implement `RemoveTranslationSlot` for HttpCache.
- [x] Add unit test for new Slot.
- [x] Perform negative (no Slot) test to confirm this solution works in real env as expected.
- [x] Get feedback on `composer.json` changes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "minimum-stability": "stable",
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.7",
+        "ezsystems/ezpublish-kernel": "^6.7@dev",
         "friendsofsymfony/http-cache-bundle": "~1.2|^1.3.8",
         "symfony/symfony": "^2.7 | ^3.1"
     },
--- a/src/SignalSlot/RemoveTranslationSlot.php
+++ b/src/SignalSlot/RemoveTranslationSlot.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\SignalSlot;
+
+use EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface;
+use eZ\Publish\Core\SignalSlot\Signal;
+use eZ\Publish\SPI\Persistence\Content\Location\Handler as SPILocationHandler;
+
+class RemoveTranslationSlot extends AbstractContentSlot
+{
+    /**
+     * @var \eZ\Publish\SPI\Persistence\Handler
+     */
+    protected $locationHandler;
+
+    /**
+     * @param \EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface $purgeClient
+     * @param \eZ\Publish\SPI\Persistence\Content\Location\Handler $spiLocationHandler
+     */
+    public function __construct(
+        PurgeClientInterface $purgeClient,
+        SPILocationHandler $spiLocationHandler
+    ) {
+        parent::__construct($purgeClient);
+        $this->locationHandler = $spiLocationHandler;
+    }
+
+    /**
+     * Checks if $signal is supported by this handler.
+     *
+     * @param \eZ\Publish\Core\SignalSlot\Signal $signal
+     *
+     * @return bool
+     */
+    protected function supports(Signal $signal)
+    {
+        return $signal instanceof Signal\ContentService\RemoveTranslationSignal;
+    }
+
+    /**
+     * Generate tags for content, relation, locations and parent locations.
+     *
+     * RemoveTranslationSignal doesn't provide locationId, so Locations need to be fetched
+     *
+     * @param \eZ\Publish\Core\SignalSlot\Signal $signal
+     *
+     * @return array
+     */
+    protected function generateTags(Signal $signal)
+    {
+        // aligned with PublishVersionSlot as translation removal is essentially publishing new Version.
+        $tags = parent::generateTags($signal);
+        foreach ($this->locationHandler->loadLocationsByContent($signal->contentId) as $location) {
+            // self
+            $tags[] = 'location-' . $location->id;
+            // children
+            $tags[] = 'parent-' . $location->id;
+            // parent
+            $tags[] = 'location-' . $location->parentId;
+            // siblings
+            $tags[] = 'parent-' . $location->parentId;
+        }
+
+        return $tags;
+    }
+}
--- a/tests/SignalSlot/RemoveTranslationSlotTest.php
+++ b/tests/SignalSlot/RemoveTranslationSlotTest.php
@@ -24,6 +24,16 @@ class RemoveTranslationSlotTest extends AbstractContentSlotTest
      */
     private $locationHandlerMock;
 
+    /**
+     * Check if required signal exists due to BC.
+     */
+    public static function setUpBeforeClass()
+    {
+        if (!class_exists(RemoveTranslationSignal::class)) {
+            self::markTestSkipped('RemoveTranslationSignal does not exist');
+        }
+    }
+
     public function setUp()
     {
         $this->locationHandlerMock = $this->getMock(LocationHandler::class);
--- a/tests/SignalSlot/RemoveTranslationSlotTest.php
+++ b/tests/SignalSlot/RemoveTranslationSlotTest.php
@@ -1,0 +1,96 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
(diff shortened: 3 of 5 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "minimum-stability": "stable",
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.7",
+        "ezsystems/ezpublish-kernel": "^6.7@dev",
```
alongosz: I'm not sure if this is the right place or the right way to require `dev-master` for `ezpublish-kernel` (took that from `ezplatform` repository), but I need this as translation removal is a part of unreleased `1.11` / `6.11`.

tests/SignalSlot/RemoveTranslationSlotTest.php
```diff
@@ -0,0 +1,111 @@
...
+     * @var \EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface|\PHPUnit_Framework_MockObject_MockObject
+     */
+    protected $purgeClientMock;
+
+    private $signal;
+
+    public function setUp()
+    {
+        $this->persistenceHandlerMock = $this->getMock(PersistenceHandler::class);
+        $this->purgeClientMock = $this->getMock(PurgeClientInterface::class);
```
alongosz: Parent `setUp` mocks deprecated interface, so I've overridden the whole method.

src/SignalSlot/RemoveTranslationSlot.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    protected $persistenceHandler;
+
+    /**
+     * @param \EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface $purgeClient
+     * @param \eZ\Publish\SPI\Persistence\Handler $persistenceHandler
+     */
+    public function __construct(
+        PurgeClientInterface $purgeClient,
+        PersistenceHandler $persistenceHandler
```
alongosz: `RemoveTranslationSignal` does not carry information about locations, so I had do make dependency on `PersistenceHandler`. Not sure if this is the optimal way, though. Any POV?
Just in case pushed whole location fetching as a fixup: 21caa86 so can be easily removed.

alongosz:
Tested manually with eZ Platform.
Review ping @andrerom @Nattfarinn (can't request a review for this repo :) )

composer.json
andrerom: This is ok, we do the same in solr: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/composer.json#L15

src/SignalSlot/RemoveTranslationSlot.php
andrerom: why do you need to clear location id? you can just clear by content id, and it will clear all relevant cache items. This is one of the many points with multi tagging, you don't need to always refer to cache by location id.

andrerom: But ok for parent I see the point, but unsure if it is needed.

tests/SignalSlot/RemoveTranslationSlotTest.php
andrerom: why not just fix parent? Assuming it is the purger the code uses.

src/SignalSlot/RemoveTranslationSlot.php
```diff
@@ -0,0 +1,76 @@
...
+            $tags[] = 'content-' . $signal->contentId;
+            // reverse relations
+            $tags[] = 'relation-' . $signal->contentId;
+        }
+
+        $locations = $this->persistenceHandler->locationHandler()->loadLocationsByContent(
+            $signal->contentId
+        );
+        foreach ($locations as $location) {
+            $tags[] = 'location-' . $location->id;
```
alongosz: > why do you need to clear location id? you can just clear by content id, and it will clear all relevant cache items. This is one of the many points with multi tagging, you don't need to always refer to cache by location id.

Interesting thing, the moment I comment-out this line, it stopped working returning cached version (checked using [this test](https://github.com/ezsystems/ezpublish-kernel/pull/2065/commits/011206b414f2cbae600533f069701da841159c54)). Don't really know anything about tagging to tell you why :(

> But ok for parent/siblings I see the point, but unsure if it is needed.

However I can't think of - looking at [REST API v2 spec](https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/rest/REST-API-V2.rst) - any case when we can access Version via parent location. That would be a good case to check if `parent-` tag is needed there.

src/SignalSlot/RemoveTranslationSlot.php
alongosz: Answered [here](#r131672666) - comment on relevant part of the code

tests/SignalSlot/RemoveTranslationSlotTest.php
alongosz: I expected all classes inheriting from parent to require fixes, but this apparently is not the case, so fixed in 0075477.

alongosz:
I just noticed I can use directly `LocationHandler` instead of `PersistenceHandler` as it is done in `PublishVersionSlot`.

In 9db1fe9 I updated `RemoveTranslationSlot` which optimizes performance. I also removed 
`'parent-'` tag as I can't think of any case it's going to be relevant for this Slot.

However without `'location-'` Slot won't work properly. Judging from other Content and Version changing Slots like `DeleteContentSlot` or `PublishVersionSlot` location is needed there.

Mentioned changes allowed me to simplify mocks, updated in a2de543.

andrerom:
> I just noticed I can use directly LocationHandler instead of PersistenceHandler as it is done in PublishVersionSlot.

Ok, if it is needed there it will be needed here also, so keep it aligned with that given this effectively is a re-publishing of the content.

alongosz:
> Ok, if it is needed there it will be needed here also, so keep it aligned with that given this effectively is a re-publishing of the content.

Perfectly aligned with `PublishVersionSlot` in 7b201b3 fixup, so there's no doubt about reasons ;) 
Also added inline comment to indicate that.
Tested with eZ Platform on `master`, `1.10`, `1.11`

Let me know when I can cleanup commits after review.

Review reping @Nattfarinn :P

alongosz:
Squashed fixups after reviews.

tests/SignalSlot/RemoveTranslationSlotTest.php
```diff
@@ -0,0 +1,96 @@
...
+                    ),
+                ]
+            );
+
+        parent::testReceivePurgesCacheForTags($signal);
+    }
+
+    public function createSignal()
+    {
+        return new RemoveTranslationSignal(
```
andrerom: Oh, shouldn't we skip the test if class is not present @alongosz? otherwise it will fail on prev branches right?

tests/SignalSlot/RemoveTranslationSlotTest.php
```diff
@@ -0,0 +1,106 @@
...
+    private $locationHandlerMock;
+
+    /**
+     * Check if required signal exists due to BC.
+     */
+    public static function setUpBeforeClass()
+    {
+        if (!class_exists(RemoveTranslationSignal::class)) {
+            self::markTestSkipped('RemoveTranslationSignal does not exist');
+        }
```
alongosz: > Oh, shouldn't we skip the test if class is not present @alongosz? otherwise it will fail on prev branches right?

Since this is unit test run directly from this package, composer install will always install newest kernel, so this shouldn't really happen. But just to be safe and consistent (we require ezpublish-kernel `^6.7`) I added this skip in 83c47b0.

Notes:
- done in `setUpBeforeClass` because data providers also rely on some usages of the Signal.
- I've checked on 6.7.x that if this class does not exist `class_exists(RemoveTranslationSignal::class)` returns false without warnings. This is ["documented" behavior](https://www.reddit.com/r/lolphp/comments/6fwu28/classnameclass_is_valid_for_any_identifier/) :laughing: 
Buf if you'd feel safer with string FQCN, I can fix that :)

ping @andrerom @Nattfarinn

tests/SignalSlot/RemoveTranslationSlotTest.php
alongosz: Ok, Added 83c47b0 fixup. For more details please see [this comment](#discussion_r132164390).

tests/SignalSlot/RemoveTranslationSlotTest.php
andrerom: > Buf if you'd feel safer with string FQCN, I can fix that :)

no, it's fine like this, it's how it is supposed to work actually. Use is not an import :)

PR #14, opened 2017-09-22
Use PHPUnit v5.7

MarioBlazek:
This PR resolves:
* `phpunit\phpunit` constraint now requires `^5.7`
* all test cases updated to use namespaced `TestCase`
* removed all deprecated `getMock` method calls
* mocks now use PHP class name constants

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,8 @@
         "symfony/symfony": "^2.7 | ^3.1"
     },
     "require-dev": {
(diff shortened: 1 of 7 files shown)


MarioBlazek:
ping @andrerom

MarioBlazek:
@andrerom wait to see the one for `ezpublish-kernel` :)

PR #15, opened 2017-10-26
[EZP-28145] Remove core http cache clearer

wizhippo:
https://jira.ez.no/browse/EZP-28145

Todo:
- [x] Update specs
- [ ] _(by someone)_ Adjust Legacy Bridge to not rely on `ezpublish.http_cache.purge` service

### DIFF
--- a/src/DependencyInjection/Compiler/KernelPass.php
+++ b/src/DependencyInjection/Compiler/KernelPass.php
@@ -20,11 +20,21 @@ class KernelPass implements CompilerPassInterface
         foreach ($container->getDefinitions() as $id => $definition) {
             if ($this->isSignalSlot($id) ||
                 $this->isSmartCacheListener($id) ||
-                $this->isResponseCacheListener($id)
+                $this->isResponseCacheListener($id) ||
+                $this->isCachePurger($id)
             ) {
                 $container->removeDefinition($id);
             }
         }
+        $container->removeAlias('ezpublish.http_cache.purger');
+        $arguments = $container->getDefinition('cache_clearer')->getArguments();
+        $arguments[0] = array_filter($arguments[0], function ($argument) {
+            if ($this->isCachePurger($argument)) {
+                return false;
+            }
+            return true;
+        });
+        $container->getDefinition('cache_clearer')->setArguments($arguments);
     }
 
     /**
@@ -55,5 +65,15 @@ class KernelPass implements CompilerPassInterface
     protected function isResponseCacheListener($id)
     {
         return $id === 'ezpublish.view.cache_response_listener';
+    }
+
+    /**
+     * @param string $id
+     *
+     * @return bool
+     */
+    protected function isCachePurger($id)
+    {
+        return strpos($id, 'ezpublish.http_cache.purger.') === 0;
     }
 }
--- a/src/DependencyInjection/Compiler/KernelPass.php
+++ b/src/DependencyInjection/Compiler/KernelPass.php
@@ -28,12 +28,12 @@ class KernelPass implements CompilerPassInterface
         }
         $container->removeAlias('ezpublish.http_cache.purger');
         $arguments = $container->getDefinition('cache_clearer')->getArguments();
-        $arguments[0] = array_filter($arguments[0], function ($argument) {
+        $arguments[0] =  array_values(array_filter($arguments[0], function ($argument) {
             if ($this->isCachePurger($argument)) {
                 return false;
             }
             return true;
-        });
+        }));
         $container->getDefinition('cache_clearer')->setArguments($arguments);
     }
 
--- a/spec/DependencyInjection/Compiler/KernelPassSpec.php
+++ b/spec/DependencyInjection/Compiler/KernelPassSpec.php
@@ -13,6 +13,7 @@ class KernelPassSpec extends ObjectBehavior
     {
         $container->getDefinitions()->willReturn([]);
         $container->getAlias('ezpublish.http_cache.purge_client')->willReturn('some_random_id');
+        $container->getAlias('ezpublish.http_cache.purger')->willReturn('some_random_id');
     }
 
     function it_is_initializable()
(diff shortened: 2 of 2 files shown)


emodric:
Does this conflict with Legacy Bridge which still uses `ezpublish.http_cache.purger`?

https://github.com/ezsystems/LegacyBridge/blob/master/bundle/Resources/config/services.yml#L260-L262

wizhippo:
@emodric Yes it would.  That service gets removed.  Not sure how to handle that one.

emodric:
Just a thought. Can't we just remove the `cache_clearer` tag from `ezplatform.http_cache.purge_client` instead of removing the alias altogether?

wizhippo:
@emodric  No as that pass has already run before this and has been gathered in the cache_clearer collection. We could leave the alias but I would rather remove it to make sure it is not used somewhere else that we have missed.

wizhippo:
@andrerom Not sure if the legacy bridge compiler pass belongs in the LegacyBundle or if is should be manged by this bundle.

Example pass https://gist.github.com/wizhippo/89f158539e5dbb513688f881edb31dea

The pass could just as easily remain in this bundle and test for `ezpublish_legacy.switchable_http_cache_purger` instead.

Ideas?

andrerom:
I'm ok with that, but then let's merge this and handle that somewhere afterwards 👍

emodric:
legacy bundle looks like a right place, yes :)

emodric:
Not sure if it's related, but just that's its out there: Once I activate the bundle, I get the following exception:

```
[2017-10-30 17:18:48] php.CRITICAL: Type error: Too few arguments to function eZ\Publish\Core\MVC\Symfony\Cache\Http\InstantCachePurger::__construct(), 1 passed in /var/www/html/netgensite/app/cache/dev/appDevDebugProjectContainer.php on line 4790 and exactly 3 expected {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Type error: Too few arguments to function eZ\\Publish\\Core\\MVC\\Symfony\\Cache\\Http\\InstantCachePurger::__construct(), 1 passed in /var/www/html/netgensite/app/cache/dev/appDevDebugProjectContainer.php on line 4790 and exactly 3 expected at /var/www/html/netgensite/vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/MVC/Symfony/Cache/Http/InstantCachePurger.php:40)"} []
```

andrerom:
@emodric Symfony 3.4? @alongosz commented about issue with that at least.

emodric:
3.3 actually. Can you point me to the comment, I wasn't able to locate it?

src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -20,11 +20,21 @@ public function process(ContainerBuilder $container)
...
-                $this->isResponseCacheListener($id)
+                $this->isResponseCacheListener($id) ||
+                $this->isCachePurger($id)
             ) {
                 $container->removeDefinition($id);
             }
         }
+        $container->removeAlias('ezpublish.http_cache.purger');
+        $arguments = $container->getDefinition('cache_clearer')->getArguments();
+        $arguments[0] =  array_values(array_filter($arguments[0], function ($argument) {
```
alongosz: In Symfony 3.4 `cache_clearer` [service definition changed](https://github.com/symfony/symfony/commit/fa62e5068e81c65cd2676675da4caeec438a33cc#diff-9301964a7acd200e8134fed55aede310R43) and now `$arguments[0]` is an instance of `Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument`
so during cache warmup we get:
```
In KernelPass.php line 36:
                                                                         
  Warning: array_filter() expects parameter 1 to be array, object given
```
Some nasty `if` checking whether this is an array could solve this, but I have a feeling there's bigger issue here. I can't see direct reference to this change in Symfony upgrade notes, so maybe we're not supposed to "mess" with this service like that?

alongosz:
@emodric I've posted it internally for no apparent reason, sorry :sweat_smile:  
Pasted [here](#pullrequestreview-72932545). This is rather a different issue.

src/DependencyInjection/Compiler/KernelPass.php
wizhippo: I like how it now removes the tagged services automatically if the service definition is removed.  But in older kernels this is not the case and the tagged service is collected and added to `cache_clearers` by the framework bundle before this compiler pass runs.  Not sure how else to handle this.

andrerom: In worst case detect symfony version with https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Kernel.php#L67

wizhippo: How about a simple `is_array`?  https://github.com/ezsystems/ezplatform-http-cache/pull/17

andrerom:
So back to @emodric's error, any idea where that comes from?

I would guess it comes from change in Legacy bridge @emodric, right?

SIDE: We also might need to refactor this part of legacy bridge soon as InstantCachePurger is deprecated: https://github.com/ezsystems/ezpublish-kernel/blob/1bdad65810d562c8a07be128f1eeb79ec1a5cfa3/eZ/Publish/Core/MVC/Symfony/Cache/Http/InstantCachePurger.php#L21

emodric:
So back to @emodric's error, any idea where that comes from?
I would guess it comes from change in Legacy bridge @emodric, right?

> Possibly, I'll test some more.

> SIDE: We also might need to refactor this part of legacy bridge soon as InstantCachePurger is deprecated: https://github.com/ezsystems/ezpublish-kernel/blob/1bdad65810d562c8a07be128f1eeb79ec1a5cfa3/eZ/Publish/Core/MVC/Symfony/Cache/Http/InstantCachePurger.php#L21

@andrerom Already tried doing that, but as far as I remember, currently not possible on 6.x.

andrerom:
> @andrerom Already tried doing that, but as far as I remember, currently not possible on 6.x.

yeah we might need to adjust some things here and there.
We'll basically somehow need to let legacy here trigger same logic signal slots does in ezplatform-http-cache, however ideally abstracted in a way that can be compatible with `purgeForCotnent`

emodric:
Yep, `purgeForContent` was the main issue :)

BTW, my issue is fixed with https://github.com/ezsystems/LegacyBridge/pull/121

PR #17, opened 2017-10-31
Add BC fix for EZP-28145 and Symfony < 3.4

wizhippo:


src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -28,13 +28,17 @@ public function process(ContainerBuilder $container)
...
-        $arguments[0] =  array_values(array_filter($arguments[0], function ($argument) {
-            if ($this->isCachePurger($argument)) {
-                return false;
-            }
-            return true;
-        }));
-        $container->getDefinition('cache_clearer')->setArguments($arguments);
+
+        // BC Symfony < 3.4
+        if (is_array($arguments[0])) {
```
alongosz: Ok, but let's keep early return convention, so
```php
if (!is_array($arguments[0])) {
    return;
}
```

alongosz: nvm :)

alongosz:
:+1:

src/DependencyInjection/Compiler/KernelPass.php
andrerom: @alongosz could you review e5f73817ca328669735b279a76ada487b6c03409

alongosz: e5f7381 approved ;)

PR #19, opened 2017-11-02
EZEE-1780: Plugin support

vidarl:
> Issue: https://jira.ez.no/browse/EZEE-1780

This PR makes it possible to extend http cache in 3rd party bundles

Depends on:
 - ~~https://github.com/ezsystems/ezpublish-kernel/pull/2136~~ _(merged)_


src/Resources/config/view_cache.yml
```diff
@@ -41,3 +50,8 @@ services:
         class: EzSystems\PlatformHttpCacheBundle\ResponseTagger\Value\LocationTagger
         tags:
             - {name: ezplatform.cache_response_tagger}
+
+    ezplatform.view_cache.response_tagger.restlocationbc:
+        class: EzSystems\PlatformHttpCacheBundle\ResponseTagger\Value\RestLocationBCTagger
+        tags:
+            - {name: ezplatform.cache_response_tagger}
```
vidarl: I really don't get the reasoning behind splitting the response taggers in delegators and values. It just creates overhead? Also, given that we keep this split, why does also the value tagges have the `ezplatform.cache_response_tagger` tag? It means they are both invoked by the dispatcher directly, and by their delicator

src/EventSubscriber/HttpCacheResponseSubscriber.php
```diff
@@ -39,9 +39,13 @@ public static function getSubscribedEvents()
 
     public function configureCache(FilterResponseEvent $event)
     {
-        $view = $event->getRequest()->attributes->get('view');
-        if (!$view instanceof CachableView || !$view->isCacheEnabled()) {
-            return;
+        if ($event->getRequest()->attributes->has('is_rest_request') && $event->getRequest()->attributes->get('is_rest_request') === true) {
```
andrerom: Maybe needs a new name, but this subscriber to me is for ContentView.

Not all rest requests are supposed to be cacheable afaik, so if (we make it) expose RestValue or something on Response _(is it maybe still on the getControllerResult?)_, then we can have a dedicated subscriber for rest responses.

src/ResponseTagger/Delegator/RestXLocationBCTagger.php
```diff
@@ -0,0 +1,33 @@
...
+    public function __construct(ResponseTagger $restLocationBCTagger)
+    {
+        $this->restLocationBCTagger = $restLocationBCTagger;
+    }
+
+    public function tag(ResponseCacheConfigurator $configurator, Response $response, $value)
+    {
+        // Rest API sets the X-Location-Id header. This needs to be changed according to http cache in use
+        if ($value === null && $response->headers->has('X-Location-Id')) {
+            $this->restLocationBCTagger->tag($configurator, $response, ['X-Location-Id' => $response->headers->get('X-Location-Id')]);
```
andrerom: Kind of duplicates #21, 

It seems a bit strange to use several layers of response taggers to map X-Location-id to xkey and remove it.

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -14,7 +14,7 @@
  * Only support BAN requests on purpose, to be able to invalidate cache for a
```
Plopix: `BAN` not updated right?

src/DependencyInjection/Compiler/PurgeTypeSettingsPass.php
```diff
@@ -0,0 +1,57 @@
...
+            throw new \InvalidArgumentException("No driver found being able to handle purge_type '$purgeType'.");
+        }
+
+        $container->setAlias('ezplatform.http_cache.purge_client', $purgeServiceId);
+
+        // If purge_type is not 'local' we must override ezpublish.http_cache.purge_client.fos too, which still uses
+        // eZ\Publish\Core\MVC\Symfony\Cache\Http\FOSPurgeClient
+        // Or else, that deprecated class is still used by cache:clear
+        if ($purgeType !== 'local') {
+            $container->setAlias('ezpublish.http_cache.purge_client', $purgeServiceId);
```
Plopix: should use *ezpublish.http_cache.purge_client* I think

src/DependencyInjection/Compiler/PurgeTypeSettingsPass.php
```diff
@@ -0,0 +1,57 @@
...
+namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class PurgeTypeSettingsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $purgeType = $container->getParameter('purge_type');
```
Plopix: we should not use the parameter here... that could be renamed. we should use the ezpublish.htt_cache.purge_type settings

src/DependencyInjection/Compiler/PurgeTypeSettingsPass.php
vidarl: No, I meant what I wrote here....
This line ensured that correct purge client was used when running `php app/console clear:cache`

However, it seems like it was decided in https://github.com/ezsystems/ezplatform-http-cache/pull/15 that we should not clear http cache when doing a clear cache

So, if https://github.com/ezsystems/ezplatform-http-cache/pull/15 stays, this line can be removed

andrerom: 15 was not necessarily the final say on it, it was to solve a bug where kernel purge system will be called on cache:clear, when if we need such logic should have the proper config for it here.

We don't clear Stash / Symfony cache either on cache:clear by default _(unless file system cache is used)_, but it should be possible somehow to do it when content or schema for some reason changes on deployment.

andrerom: But this line can indeed be removed, ca:cl is not supposed to call kernel purge client anymore with this bundle enabled. Actually almost all http cache code in kernel is deprecated in favor of this bundle.

Plopix: in any case are we sure we don't want to use `ezplatform.http_cache.purge_client` and not `ezpublish.http_cache.purge_client` ?

vidarl: I don't think I understood your last comment, but I have removed the line in question as it does not do anything after https://github.com/ezsystems/ezplatform-http-cache/pull/15 was merged

src/EventSubscriber/HttpCacheResponseSubscriber.php
andrerom: see #25 I would think this part is not needed anymore if we handle `X-Location-Id` issue in separate listener

src/DependencyInjection/Compiler/PurgeTypeSettingsPass.php
vidarl: Hmm.. Not sure that is possible in a compiler pass. You know how?

Plopix: Not by default, I think, no way :( But for sure we should do that on `purge_type`.

You will have to set the value as a parameter here:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php#L392
```php
$container->setParameter("ezpublish.http_cache.purge_type", $config['http_cache']['purge_type']);
```

And then refer to it 
```php
$purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
```

Then comes the SiteAccess Awareness, but I think that is global then a parameter would be fine.

@andrerom  @bdunogier  another idea?

src/EventSubscriber/HttpCacheResponseSubscriber.php
Plopix: +1

src/ResponseTagger/Delegator/RestXLocationBCTagger.php
Plopix: and now it is also useless because of #25 I think

src/DependencyInjection/Compiler/PurgeTypeSettingsPass.php
vidarl: Right, fixed as suggested

src/PurgeClient/VarnishPurgeClient.php
vidarl: yeah, that was an old comment... Removed it

src/PurgeClient/PurgeClientInterface.php
```diff
@@ -16,11 +16,16 @@
 interface PurgeClientInterface extends KernelPurgeClientInterface
 {
     /**
-     * Triggers the cache purge $locationIds.
+     * Triggers the cache purge of $tags.
      *
      * It's up to the implementor to decide whether to purge $locationIds right away or to delegate to a separate process.
```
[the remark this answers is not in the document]
vidarl: So you insist on calling it @locationIds ?
It is not consistent with https://github.com/ezsystems/ezplatform-http-cache/pull/19/files#diff-cdb40d122691729c63752db68cddf3f9R31 and other places
And besides, there is not only location ids, also path-XX, content-XX and content-type-XX etc.....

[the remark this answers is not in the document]
andrerom: No, you misunderstand me. I was referring to text here referring to `$locationIds` when it should be aligned with the other changes you make to refer to `tags` instead.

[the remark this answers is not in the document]
vidarl: ah... missed that one.. will fix

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
+# Driver support
+
+You may add integration with other http caches using the extension points provided by this bundle.
+
+The following extension points are available
+ - PurgeClient
+ - TagHandler
+ - FOS TagHandler
+
+If you write a new PurgeClient driver, you **must** also create a corresponding TagHandler and vise
```
DominikaK: ...`TagHandler and vice`

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
...
+versa. Creating a FOS TagHandler is optional.
+
+
+## PurgeClient
+
+The PurgeClient is responsible for sending purge requests to the http cache when content is about to be invalidated.
+The PurgeClient must implement EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface and can be registered
+with the following code in service.yml:
```
DominikaK: Isn't it service**S**.yml?

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
...
+            - {name: ezplatform.http_cache.tag_handler, purge_type: myhttpcache}
+
+```
+
+Any service which implements the TagHandlerInterface must be tagged with `ezplatform.http_cache.tag_handler` in order
+to be registered as such.
+
+## FOS TagHandler
+
+The FOS Http cache bundle also have a TagHandler which is not used by eZ Platform except for one thing, the
```
DominikaK: ...`also has a TagHandler`...

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
...
+`fos:httpcache:invalidate:tag` command. With this command you may explicit invalidate cache by tag.
```
DominikaK: ...`you may explicitly invalidate cache`...

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
...
+
+Normally, you would not need to implement your own FOS TagHandler as the ezplatform-http-cache bundle ships with a
+default one which uses the PurgeClient to invalidate the given tags.
+If you anyway need to write your own FOS TagHandler, you may register it with the following code in service.yml:
```
DominikaK: `If you need to write your own FOS TagHandler anyway`...

docs/drivers.md
vidarl: yes, it is. good catch....

docs/drivers.md
vidarl: fixed

docs/drivers.md
vidarl: fixed

docs/drivers.md
vidarl: fixed

docs/drivers.md
vidarl: fixed

src/DependencyInjection/EzSystemsPlatformHttpCacheExtension.php
```diff
@@ -35,4 +35,9 @@ public function prepend(ContainerBuilder $container)
         $container->prependExtensionConfig('fos_http_cache', $config);
         $container->addResource(new FileResource($configFile));
     }
+
+    public function getAlias()
+    {
+        return 'ez_platform_cache';
```
andrerom: `ez_platform_http_cache`, method can afaik be avoided by renaming this class to EzPlatformHttpCacheExtension which `getContainerExtension()` will then need to be adapted for.

vidarl: I don't think that will be better. If I rename class, then I think the most clean way should be to also override Bundle::getContainerExtensionClass() ( and not write logic concerning class name in getContainerExtension() ).

So IMO it is either to create a getAlias() or a getContainerExtensionClass() method.

vidarl: Replaced getAlias() with getContainerExtensionClass()

composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "minimum-stability": "stable",
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.7@dev || ^7.0@dev",
+        "ezsystems/ezpublish-kernel": "~6.7.7 || ^6.12.1 || ^7.0",
```
andrerom: you'll need to keep dev for tests to continue to work

andrerom: fixed in 358a682

docs/drivers.md
```diff
@@ -0,0 +1,66 @@
...
+
+The PurgeClient is responsible for sending purge requests to the http cache when content is about to be invalidated.
+The PurgeClient must implement EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface and can be registered
+with the following code in services.yml:
+
+```
+services:
+    ezplatform.http_cache_myhttpcachebundle.purge_client.myhttpcache:
+        class: EzSystems\PlatformMyHttpCacheBundle\PurgeClient\MyHttpCachePurgeClient
+        arguments: ["@ezplatform.http_cache.cache_manager"]
```
alongosz: Nitpick: Let's advocate in doc single quotes as well here. Sorry to be persistent and boring, just trying to clean the mess I made some time ago ;)

src/EzSystemsPlatformHttpCacheBundle.php
```diff
@@ -15,5 +17,35 @@ public function build(ContainerBuilder $container)
...
+                $this->extension = $extension;
+            } else {
+                $this->extension = false;
+            }
+        }
+
+        if ($this->extension) {
+            return $this->extension;
+        }
+    }
```
alongosz: Why can't we just `return new EzPlatformHttpCacheExtension()`? See [Symfony doc](https://symfony.com/doc/current/bundles/extension.html#manually-registering-an-extension-class)

src/DependencyInjection/Compiler/DriverPass.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class DriverPass implements CompilerPassInterface
```
alongosz: Nitpick: Please add brief description on what this Compiler Pass does.

src/EzSystemsPlatformHttpCacheBundle.php
vidarl: ah.... man.... This we talked about yesterday
Ok, I'll change it again

src/DependencyInjection/Compiler/DriverPass.php
vidarl: fixed

src/EzSystemsPlatformHttpCacheBundle.php
vidarl: fixed

alongosz: Sorry, I think I missed the point of that discussion and didn't see this code yesterday. Doing too many things at the same time ;)

composer.json
andrerom: Fixed again in b8a23482497c7925fde9f844b5c39a54c74ee357

please pull in changes before rebasing ;)

PR #22, opened 2017-11-05
Purge type HTTP just does not work

Plopix:
This bundle relies on `ezplatform.http_cache.purge_client` but the code in ez kernel sets `ezpublish.http_cache.purge_client` 
Wiring has not been done, which also mean, that this feature was not tested.


Plopix:
ping @vidarl @andrerom @bdunogier 
This PR and #21 should bring back a working version with Varnish of 1.12

Plopix:
@bdunogier @andrerom @vidarl 
> The package is still experimental, as indicated in the README.

But it is enabled by default in the last stable and LTS :/ 
Should we remove it from there?

Cool this PR helped!

PR #23, opened 2017-11-06
EZP-28160: Wrong slot configuration for HTTP cache purge after copying subtree

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28160

## Description

https://github.com/ezsystems/ezplatform-http-cache/pull/18 added 
`EzSystems\PlatformHttpCacheBundle\SignalSlot\CopySubtreeSlot`, which should be connected to `LocationService\CopySubtreeSignal` instead of `ContentService\CopySubtreeSignal`. 

_I'm not sure how I've overlooked this while preparing a commit. My fault._

### DIFF
--- a/src/Resources/config/slot.yml
+++ b/src/Resources/config/slot.yml
@@ -20,7 +20,7 @@ services:
         class: EzSystems\PlatformHttpCacheBundle\SignalSlot\CopySubtreeSlot
         parent: ezplatform.http_cache.signalslot.abstract_content
         tags:
-            - { name: ezpublish.api.slot, signal: ContentService\CopySubtreeSignal }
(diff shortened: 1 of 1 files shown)


adamwojs:
Review ping @andrerom @alongosz

andrerom:
oh, yes `ContentService\CopySubtreeSignal` does indeed not exists, shoulød have seen that as well :|

adamwojs:
I think that we should cover `\EzSystems\PlatformHttpCacheBundle\DependencyInjection\EzSystemsPlatformHttpCacheExtension` with tests, to check if it registers basic slots etc. to avoid such situation in the feature. I guess it would be possible.

PR #24, opened 2017-11-06
Update Readme.md

Plopix:
Update the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,10 @@
 
 # platform-http-cache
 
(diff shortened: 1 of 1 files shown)


Plopix:
@bdunogier @andrerom 
Decision has to be made of by default and stable or not by default an experimental

Thanks! (that is important to us, in case of migration etc.)

andrerom:
Merging as I'm about to open larger changes on doc.

PR #25, opened 2017-11-07
EZP-28183: Requests (incl from REST) with X-Location-Id not tagged

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28183
>     REST request are not properly tagged and hence not purged as they should, custom controllers not using ContentView is also affected.
> Replaces #21 

@bdunogier @Plopix : Here is somewhat what I had in mind here to begin with. This approach does not need to use response tagger in a way it was made for, and does not need to force caching of all REST responses. It just does one thing, transform the old value to the new one, on top of that it triggers deprecation warning pointing to doc we will need to write on how to deal with tags... (:

Todo:
- [x] Spec
- _(separate pr)_ Doc on using tags

### DIFF
--- a/src/EventSubscriber/XLocationIdResponseSubscriber.php
+++ b/src/EventSubscriber/XLocationIdResponseSubscriber.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Rewrites the X-Location-Id HTTP header.
+ *
+ * This is a BC layer for custom controllers (including REST server) still
+ * using X-Location-Id header which is now deprecated. For
+ * full value of tagging, see docs/using_tags.md for how to take advantage of the
+ * system.
+ */
+class XLocationIdResponseSubscriber implements EventSubscriberInterface
+{
+    const LOCATION_ID_HEADER = 'X-Location-Id';
+
+    /**
+     * @var string
+     */
+    private $tagHeader;
+
+    public function __construct($tagHeader)
+    {
+        $this->tagHeader = $tagHeader;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [KernelEvents::RESPONSE => ['rewriteCacheHeader', -5]];
+    }
+
+    public function rewriteCacheHeader(FilterResponseEvent $event)
+    {
+        $response = $event->getResponse();
+        if (!$response->headers->has(static::LOCATION_ID_HEADER)) {
+            return;
+        }
+
+        @trigger_error(
+            'X-Location-Id is no longer preferred way to tag content responses, see ezplatform-http-cache/docs/using_tags.md',
+            E_USER_DEPRECATED
+        );
+
+        // Map the tag, even if not officially supported, handle comma separated values as was possible with Varnish
+        $tags = array_map(
+            function ($id) {
+                return 'location-' . trim($id);
+            },
+            explode(',', $response->headers->get(static::LOCATION_ID_HEADER))
+        );
+
+        if ($response->headers->has($this->tagHeader)) {
+            $tags = array_merge($response->headers->get($this->tagHeader, null, false) , $tags);
+        }
+
+        // @todo we need to use abstract tag writer to also be able to support Fastly
+        // FOS has some stuff around this in 2.x but not in a good way in 1.x
+        $response->headers->set($this->tagHeader, array_unique($tags));
+        $response->headers->remove(static::LOCATION_ID_HEADER);
+    }
+}
--- a/src/Resources/config/view_cache.yml
+++ b/src/Resources/config/view_cache.yml
@@ -7,6 +7,12 @@ services:
         tags:
             - { name: kernel.event_subscriber }
 
+    ezplatform.x_location_id.response_subscriber:
+        class: EzSystems\PlatformHttpCacheBundle\EventSubscriber\XLocationIdResponseSubscriber
+        arguments: ['%ezplatform.http_cache.tags.header%']
+        tags:
+            - { name: kernel.event_subscriber }
+
     ezplatform.view_cache.response_configurator:
         class: EzSystems\PlatformHttpCacheBundle\ResponseConfigurator\ConfigurableResponseCacheConfigurator
         arguments:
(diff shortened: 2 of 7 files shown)


src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+    public function rewriteCacheHeader(FilterResponseEvent $event)
+    {
+        $response = $event->getResponse();
+        if (!$response->headers->has('X-Location-Id')) {
+            return;
+        }
+
+        if ($this->logger instanceof LoggerInterface) {
+            $this->logger->notice(
+                'X-Location-Id rewritten to use ezplatform-http-cache tag header, see \<bundle\>/docs/using_tags.md for more info'
```
bdunogier: What about a good old `E_USER_DEPRECATED` rather than a notice ? Those are way more visible by developers. As long as our own code doesn't throw it ;)

src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+            function ($id) {return 'location-' . trim($id);},
+            explode(',', $response->headers->get('X-Location-Id'))
+        );
+
+        // Check if there is already exiting tags and merge them
+        if ($response->headers->has($this->tagHeader)) {
+            $tags += $response->headers->get($this->tagHeader, null, false);
+        }
+
+        // Set headers
```
bdunogier: I know this is WIP but I really wouldn't leave such comments :-)

Also applies to the comments above. If the statements aren't readable (the `->has()` one is readable, in my opinion, for instance), private functions can be used.

src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+        // @todo we need to use abstract tag writer to also be able to support Fastly
```
bdunogier: What does the abstract tag writer do that we don't do here ? Since we use the header name as set in the fos configuration, it should work with Fastly, should it not ?

src/Resources/config/view_cache.yml
```diff
@@ -7,6 +7,14 @@ services:
         tags:
             - { name: kernel.event_subscriber }
 
+    ezplatform.x_location_id.response_subscriber:
+        class: EzSystems\PlatformHttpCacheBundle\EventSubscriber\XLocationIdResponseSubscriber
+        arguments:
+            - '%fos_http_cache.tags.header%'
+            - '@?logger'
```
bdunogier: If we use `E_USER_DEPRECATED`, this isn't required anymore.

src/EventSubscriber/XLocationIdResponseSubscriber.php
andrerom: I meant to check with @vidarl, but the value is different in Fastly examples, it's separated by space: https://docs.fastly.com/guides/purging/getting-started-with-surrogate-keys#understanding-surrogate-keys

So if Fastly does not support what we currently do*, then we need to come up with something for abstracting this. And we might actually have to do that anyway, as if we want to promote usage of FosHttpCache TagsHandler for instance for custom tag usage, without hard-coding xkey header, it currently [always writes comma separated values](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/1.4/src/Handler/TagHandler.php#L91-L94) in 1.x _(2.x to but it has [a formater object](https://github.com/FriendsOfSymfony/FOSHttpCache/search?utf8=%E2%9C%93&q=getTagsHeaderValue&type=) you can change to something else)_.


* Respond with several headers to comply with xkey doc: https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst

andrerom:
@bdunogier Update, also added spec.

@vidarl As discussed [above](https://github.com/ezsystems/ezplatform-http-cache/pull/25#discussion_r149382449), could you comment on if Fastly is ok with the separated headers approach we current take here or if it needs space separated values like it's doc suggests?

Plopix:
I did not follow everything but it seems that #21 is useless now.

I like this one! Can't wait to test it ;)

andrerom:
PR updated, config is using own param now instead of relying on anything from FOS it instead makes it reuse our param.

Ready for review

src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public static function getSubscribedEvents()
+    {
+        return [KernelEvents::RESPONSE => ['rewriteCacheHeader', -5]];
+    }
+
+    public function rewriteCacheHeader(FilterResponseEvent $event)
+    {
+        $response = $event->getResponse();
+        if (!$response->headers->has('X-Location-Id')) {
```
bdunogier: Nitpick: I'd make `'X-Location-Id'` a class constant.

andrerom: But it would be public 😢   Un indispensable?

bdunogier: No, it was a simple nitpick. I don't think it's an issue that the constant is public, but if it itches your goat, you can skip it :)

andrerom: 🐐

andrerom: On the other side, in this case it will never and can never change, so actually should be fine even for my 🐐  So will fix that on merge :)

andrerom:
Merged with 🐐 's and all , thanks all.

Plopix:
thx @andrerom tested and it works well!

src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -0,0 +1,68 @@
...
+            explode(',', $response->headers->get(static::LOCATION_ID_HEADER))
+        );
+
+        if ($response->headers->has($this->tagHeader)) {
+            $tags = array_merge($response->headers->get($this->tagHeader, null, false) , $tags);
+        }
+
+        // @todo we need to use abstract tag writer to also be able to support Fastly
+        // FOS has some stuff around this in 2.x but not in a good way in 1.x
+        $response->headers->set($this->tagHeader, array_unique($tags));
```
vidarl: I don't like this. Now we have logic for setting xkey both here and in https://github.com/ezsystems/ezplatform-http-cache/blob/a4ed9b6b8d9a9dfc4a0f92e8a1e1de56a877d9c3/src/ResponseConfigurator/ConfigurableResponseCacheConfigurator.php#L63

This should be done in one place.
I fancy to replace the whole ConfigurableResponseCacheConfigurator with FOS\HttpCacheBundle\Handler\TagHandler, but that is not quite straight forward.
So instead I am thinking just adding a TagHandler and make sure ConfigurableResponseCacheConfigurator  and XLocationIdResponseSubscriber  uses that one for setting the headers. What do you say?

vidarl: Okay, Adding our own TagHandler based on FOS\HttpCacheBundle\Handler was not a good idea for several reasons.
One was that we are not able to run it's constructor without also adding a CacheInvalidator ( kinda replaces our PurgeClient) and that it's tag handling doesn't really suit our needs in regards to either xkey nor fastly.

vidarl: I'll then create our abstraction for this

andrerom: thread marked as obsolete :)

PR #26, opened 2017-11-07
Set FOS config for tags header

andrerom:
In the need for specifying tags.header config, I copied over logic from CoreBundle.

@bdunogier Is this ok approach? I would assume we need it here anyway once stuff like this is removed from kernel.


Plopix:
+1

andrerom:
Ok, was afraid of that, then we need a pass or custom config for this ala what fos does for tag handler _(but on 1.x tag handler seems to only work if varnish is configured, so I think we need something abstract of this)_: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/FOSHttpCacheExtension.php#L401

andrerom:
@vidarl Checked this and this PR works as intended, debugging config shows they have been set.

What is wrong is that in #25 I assumed this would be available as param as well, it's not, so I need to do a pass there to extract this. There is `fos_http_cache.tag_handler.header`, however it and rest of tag handler is only set if proxy client is "varnish" or "custom" so I can not rely on it: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/DependencyInjection/FOSHttpCacheExtension.php#L331-L338

Merging this and continuing in #25 to get the right value out.

PR #29, opened 2017-11-09
Fixes #28: Missing purgeAllContent()

### LINKED ISSUE #28: EzSystems\PlatformHttpCacheBundle\Proxy\TagAwareStore::purgeAllContent() not implemented
purgeAllContent method doesn't seem to be implemented.

https://github.com/ezsystems/ezplatform-http-cache/blob/250149c9a29b7e556c05931a503280568227de87/src/Proxy/TagAwareStore.php#L142

andrerom:
Attempt to fix #28

### DIFF
--- a/src/Proxy/TagAwareStore.php
+++ b/src/Proxy/TagAwareStore.php
@@ -162,6 +162,14 @@ class TagAwareStore extends Store implements RequestAwarePurger
     }
 
     /**
+     * Purge all cache.
+     */
+    protected function purgeAllContent()
+    {
+        $this->getFilesystem()->remove((new Finder())->in($this->root));
+    }
+
+    /**
      * Purges cache for tag.
      *
      * @param string $tag
--- a/src/PurgeClient/LocalPurgeClient.php
+++ b/src/PurgeClient/LocalPurgeClient.php
@@ -55,6 +55,8 @@ class LocalPurgeClient implements PurgeClientInterface
      */
     public function purgeAll()
     {
-        $this->cacheStore->purgeAllContent();
+        $purgeRequest = Request::create('http://localhost/', 'PURGE');
+        $purgeRequest->headers->set('X-Location-Id', '*');
+        $this->cacheStore->purgeByRequest($purgeRequest);
     }
 }


src/Proxy/TagAwareStore.php
```diff
@@ -162,6 +162,14 @@ function ($id) {return 'location-' . $id;},
     }
 
     /**
+     * Purge all cache.
+     */
+    protected function purgeAllContent()
+    {
+        $this->fs->remove((new Finder())->in($this->root));
```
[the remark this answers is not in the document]
andrerom: Right, I see the setter is not used indeed.

Does legacy bridge bark at the method added being protected? Do you have a stack trace?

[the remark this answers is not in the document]
andrerom: Adjusted in 2e4dc81, you confirm this working as expected with the change you proposed?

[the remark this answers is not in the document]
volkanakb: thx, works fine

src/PurgeClient/LocalPurgeClient.php
```diff
@@ -55,6 +55,8 @@ function ($tag) {
      */
     public function purgeAll()
     {
-        $this->cacheStore->purgeAllContent();
+        $purgeRequest = Request::create('http://localhost/', 'PURGE');
```
alongosz: Hmm, doesn't actual host come from some parameter? At first glance I thought it was a test.

src/PurgeClient/LocalPurgeClient.php
```diff
@@ -55,6 +55,8 @@ function ($tag) {
...
+        $purgeRequest->headers->set('X-Location-Id', '*');
```
alongosz: Didn't we have constant for `X-Location-Id` header name? I vaguely remember some discussion about it in one of the other PRs.

src/PurgeClient/LocalPurgeClient.php
andrerom: This is same as in purge() above, which comes from https://github.com/ezsystems/ezpublish-kernel/blob/1bdad65810d562c8a07be128f1eeb79ec1a5cfa3/eZ/Publish/Core/MVC/Symfony/Cache/Http/LocalPurgeClient.php#L47

reason is host does not matter with the symfony proxy implementation, it's local in the same code base.

src/PurgeClient/LocalPurgeClient.php
andrerom: yeah, but it's not really relevant class to take constant imho, as it has a special purpose.

PR #31, opened 2017-11-10
Fixed SetEnv syntax in README.md

volkanakb:
according to https://httpd.apache.org/docs/2.4/mod/mod_env.html there should be no equals sign when defining an Env Variable


bdunogier:
Thank you for this fix !

PR #32, opened 2017-11-13
Definded handling of tag header in a separate TagHandlerInterface

vidarl:
So, instead of dealing with xkeys in both XLocationIdResponseSubscriber and ConfigurableResponseCacheConfigurator I created a interface for it, TagHandlerInterface. Not 100% happy with then name as it might be confusing as FOS also have a TagHandler.

### DIFF
--- a/src/DependencyInjection/Compiler/KernelPass.php
+++ b/src/DependencyInjection/Compiler/KernelPass.php
@@ -7,6 +7,8 @@ namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use EzSystems\PlatformHttpCacheBundle\Handler\TagHandler;
 
 /**
  * Disables some of the http-cache services declared by the kernel so that
@@ -44,6 +46,14 @@ class KernelPass implements CompilerPassInterface
 
         if ($container->getAlias('ezpublish.http_cache.purge_client') == 'ezpublish.http_cache.purge_client.fos') {
             $container->setAlias('ezplatform.http_cache.purge_client', 'ezplatform.http_cache.purge_client.fos');
+        }
+
+        $purgeType = $container->getParameter('purge_type');
+        if ($purgeType === 'http') {
+            // Injecting our own Tag handler
+            $definition = $container->getDefinition('fos_http_cache.handler.tag_handler');
+            $definition->setClass(TagHandler::class);
+            $definition->addArgument(new Reference('ezplatform.http_cache.purge_client.fos'));
         }
     }
 
--- a/src/Handler/TagHandlerInterface.php
+++ b/src/Handler/TagHandlerInterface.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace EzSystems\PlatformHttpCacheBundle\Handler;
+
+use Symfony\Component\HttpFoundation\Response;
+
+interface TagHandlerInterface
+{
+    public function addTagHeaders(Response $response, array $tags);
+}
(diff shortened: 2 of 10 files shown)


vidarl:
Please review :  @andrerom @bdunogier, @alongosz

src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -45,6 +46,14 @@ public function process(ContainerBuilder $container)
         if ($container->getAlias('ezpublish.http_cache.purge_client') == 'ezpublish.http_cache.purge_client.fos') {
             $container->setAlias('ezplatform.http_cache.purge_client', 'ezplatform.http_cache.purge_client.fos');
         }
+
+        $purgeType = $container->getParameter('purge_type');
+        if ($purgeType === 'http') {
+            // Injecting our own Tag handler
+            $definition = $container->getDefinition('fos_http_cache.handler.tag_handler');
+            $definition->setClass('EzSystems\PlatformHttpCacheBundle\Handler\TagHandler');
```
alongosz: Nitpick: please use `::class`

src/Resources/config/services.yml
```diff
@@ -24,3 +24,12 @@ services:
     ezplatform.http_cache.tag_aware_store:
         class: EzSystems\PlatformHttpCacheBundle\Proxy\TagAwareStore
         arguments: ["%ezplatform.http_cache.store.root%"]
+
+    ezplatform.http_cache.tag_handler:
+        class: EzSystems\PlatformHttpCacheBundle\Handler\TagHandler
+        # use parent: ???
```
alongosz: Hmm, looks like a TODO :)
Answer: in this case probably not as you've overridden constructor to use our custom arguments / services

src/ResponseConfigurator/ConfigurableResponseCacheConfigurator.php
```diff
@@ -32,11 +33,14 @@ class ConfigurableResponseCacheConfigurator implements ResponseCacheConfigurator
      */
     private $defaultTtl;
 
-    public function __construct($enableViewCache, $enableTtlCache, $defaultTtl)
+    private $tagHandler;
```
alongosz: Nitpick: please add a type hint in PhpDoc

src/DependencyInjection/Compiler/KernelPass.php
vidarl: Fixed

src/Resources/config/services.yml
vidarl: Yeah... That line was never meant to be committed.. Fixed.

src/ResponseConfigurator/ConfigurableResponseCacheConfigurator.php
vidarl: fixed

src/Handler/TagHandlerInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace EzSystems\PlatformHttpCacheBundle\Handler;
+
+use Symfony\Component\HttpFoundation\Response;
+
+interface TagHandlerInterface
```
andrerom: Alternative name, which also overlaps with FOS, but with 2.x, is: `ReponseTaggerInterface`
see: http://foshttpcachebundle.readthedocs.io/en/latest/features/tagging.html

Similar to what they do: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/Http/SymfonyResponseTagger.php

vidarl: Yes, I am aware of that. If there is consensus that ResponeTaggerInterface is a better name, I can change it

andrerom: It's fine keeping it like it is for now.
Can you rebase so we can merge?

vidarl: Rebased....

PR #34, opened 2017-11-20
Add support for purging several tags at once with Varnish

andrerom:
Varnish modules v0.10.2 and higher supports purging several tags at once, so use that by default.
Also allow to configure the purge header used by Varnish proxy purger


andrerom:
Putting this on hold for now, as we'll soon look into refactoring for FOSHttpCache 2.x and there we'll get things like this for free.

andrerom:
Closing, this will cause cache manager to use comma separated values which is not supported by Varnish. So for FOSHttpCache 2.x then.

PR #35, opened 2017-11-22
Exposing ezplatform.http_cache.purge_type parameter for 3rd party drivers

vidarl:
It might be needed for 3rd party drivers to acquire the purge_type setting. With this new parameter you may get it without depending on kernel

### DIFF
--- a/src/DependencyInjection/Compiler/KernelPass.php
+++ b/src/DependencyInjection/Compiler/KernelPass.php
@@ -41,6 +41,9 @@ class KernelPass implements CompilerPassInterface
             return true;
         }));
         $container->getDefinition('cache_clearer')->setArguments($arguments);
+
+        // Let's re-export purge_type setting so that driver's don't have to depend on kernel in order to acquire it
+        $container->setParameter('ezplatform.http_cache.purge_type', $container->getParameter('ezpublish.http_cache.purge_type'));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vidarl:
@andrerom, @alongosz : Please review

src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -41,6 +41,9 @@ public function process(ContainerBuilder $container)
             return true;
         }));
         $container->getDefinition('cache_clearer')->setArguments($arguments);
+
+        // Let's re-export purge_type setting so that driver's don't have to depend on kernel in order to acquire it
+        $container->setParameter('ezplatform.http_cache.purge_type', $container->getParameter('ezpublish.http_cache.purge_type'));
```
andrerom: code above is conditional 

can be fixed by moving BC code for symfony < 3.4 into own method so code added afterwards in this method are not affected.

andrerom: Also wondering if we should remove `'ezpublish.http_cache.purge_type` parameter? hmm, probably not :)

vidarl: Moved BC code into separate function.

I don't think we should remove `ezpublish.http_cache.purge_type`. It is not our (ezplatform-http-cache) business to interfere if someone choose to depends on it (and the kernel) IMO

PR #37, opened 2017-11-27
EZP-28326: Http Cache not updated when moving content on Varnish & Symfony Proxy

andrerom:
> issue: https://jira.ez.no/browse/EZP-28326

With this bundle, REST HttpCache is not cleared when moving content,
editing content types, (...) And given the time it will take to fix
REST server to return data that can be used for ResponseTaggers and
how this also affects custum responses given, as we now only purge
by specific tags. Adding this is thus safests.

This somewhat brings back what @Plopix suggested in #21, essentially making sure we
add all tags also to X-Location responses by loading location, but to be on the safe
side it does it using sudo() and catching NotFound. Opted not to reuse ReponseTaggers
here as they don't fully fit, and as we want to rewrite this header independently of
if view cache is enabled or not.

cc @mnocon

### DIFF
--- a/spec/EventSubscriber/XLocationIdResponseSubscriberSpec.php
+++ b/spec/EventSubscriber/XLocationIdResponseSubscriberSpec.php
@@ -5,8 +5,12 @@
  */
 namespace spec\EzSystems\PlatformHttpCacheBundle\EventSubscriber;
 
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\Core\Base\Exceptions\NotFoundException;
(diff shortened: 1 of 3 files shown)


bdunogier:
Given where this is fixed (the XLocationID tagger), having custom code to handle a BC case is a good solution.

andrerom:
@bdunogier They where found by QA on this one: https://jira.ez.no/browse/EZP-28183 But we can create separate one for this as it is not specific to Varnish., just did that here: https://jira.ez.no/browse/EZP-28326

PR #38, opened 2017-11-27
Fixed EZP-28243, BC break caused by EZEE-1780

vidarl:
In kernel, purge_type may be a service id. This was not taken into account in https://github.com/ezsystems/ezplatform-http-cache/pull/19

This PR fixes that issue, ref. https://jira.ez.no/browse/EZP-28243

### DIFF
--- a/src/DependencyInjection/Compiler/DriverPass.php
+++ b/src/DependencyInjection/Compiler/DriverPass.php
@@ -5,6 +5,7 @@ namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
+use eZ\Publish\Core\MVC\Symfony\Cache\PurgeClientInterface;
 
 /**
  * We support http cache drivers to be provided by 3rd party bundles.
@@ -19,7 +20,16 @@ class DriverPass implements CompilerPassInterface
         $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
         $configuredPurgeClientServiceId = static::getTaggedService($container, 'ezplatform.http_cache.purge_client');
         if ($configuredPurgeClientServiceId === null) {
-            throw new \InvalidArgumentException("No driver found being able to handle purge_type '$purgeType'.");
+            // BC Check. purge_type may also be a service name implementing PurgeClientInterface
+            $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+            if ($container->has($purgeType)) {
+                if (!$container->get($purgeType) instanceof PurgeClientInterface) {
+                    throw new \InvalidArgumentException('Invalid ezpublish.http_cache.purge_type, it needs to implement PurgeClientInterface.');
+                }
+                $configuredPurgeClientServiceId = $purgeType;
+            } else {
+                throw new \InvalidArgumentException("No driver found being able to handle purge_type '$purgeType'.");
+            }
         }
         $container->setAlias('ezplatform.http_cache.purge_client', $configuredPurgeClientServiceId);
 


vidarl:
@andrerom, @alongosz  : Please review

src/DependencyInjection/Compiler/DriverPass.php
```diff
@@ -19,7 +20,16 @@ public function process(ContainerBuilder $container)
         $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
         $configuredPurgeClientServiceId = static::getTaggedService($container, 'ezplatform.http_cache.purge_client');
         if ($configuredPurgeClientServiceId === null) {
-            throw new \InvalidArgumentException("No driver found being able to handle purge_type '$purgeType'.");
+            // BC Check. purge_type may also be a service name implementing PurgeClientInterface
+            $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+            if ($container->has($purgeType)) {
+                if (!$container->get($purgeType) instanceof PurgeClientInterface) {
```
vidarl: I here do a $container->get() inside a kernel-pass, same as https://github.com/ezsystems/ezpublish-kernel/commit/5b45826043a3c8bef1362afeede2038eb3237260#diff-469a42b65cf73f41f143b52df89d1e70R74

This might not be a good idea....Dependent on the dependency of that service, that might or might not work

andrerom: isn't it possible to aldo get just the definition and check class that way?

andrerom:
Given usage by service id was not working before 1.12.1 (not out yet), I don't think this necessarily is a BC break. But if this is ok with everyone then it's of course good to make it consistent.

src/DependencyInjection/Compiler/DriverPass.php
vidarl: Then you get class name as string ('\MyNamespace\PurgType\AwsomePurge"). AFAIK, you then would need to use reflection in order to check if it implements the interface AFAIK.
Is that the way to go?

andrerom: hmm, you might be right. I'm unsure what is the most correct here. maybe @bdunogier has input on that :)

vidarl: My two cents would be do remove support for having purge_type=[service_id] entirely, as it has always been a broken feature, and no-one have requested it to be fixed ---> no-one has ever used it + ezplatform_http_cache now has plugin support for 3rd party purgers

andrerom: I'm fine with that. But then clarify further in the Driver.md doc and double check with @mnocon if there are other places / doc that will need clarification for it _(aka I assume he got the impression to use service id from somewhere)_

adamwojs: We should be able to do this by using:

```php
$definition = $container->getDefinition('ezpublish.http_cache.purge_type');
if (!is_subclass_of($definition->getClass(), PurgeClientInterface::class)) {
    // ...
}
```

Something similar is done here: https://github.com/symfony/framework-bundle/blob/3.4/DependencyInjection/Compiler/CacheCollectorPass.php#L45

mnocon: Actually I've used purge_type tag at first (using the drivers.md doc - it was clear enough for me), but then after Vidar's suggestion: https://jira.ez.no/browse/EZP-28243?focusedCommentId=219154 I've decided to try again with `purge_type: service_id` to see if this approach works.

I think it's only mentioned in the doc once, here: https://github.com/ezsystems/ezpublish-kernel/pull/2136/files#diff-78c39a773f9efe5f9c22eca3a9f246fbR330

vidarl: so, what do we do? Fix as suggested by Adam (thanx btw), or remove possibility to use service id?

andrerom: imo doc, one way of doing things is better then many :)

andrerom:
Solved in doc here for now: https://github.com/ezsystems/ezplatform-http-cache/commit/92848da8f20650536507af4b304c8a8f57e4db0c

@vidarl maybe you and @bdunogier can improve the doc in kernel as part of the cleanup on injecting that config from bundle.

PR #39, opened 2017-11-27
EZP-28326: Use improved signals with parent id's to clear relevant cache

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28326

Makes it possible to now expire parent location as well as sibling cache for
affected content.

Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2169

Todo:
- [x] Update tests once kernel part is in


andrerom:
@adamwojs @alongosz Ready for review , 7.0 job that failed for missing signals has been restarted now that 7.0 has latests changes.

PR #41, opened 2017-12-14
Move Role Context Provider from kernel to http cache bundle

andrerom:
Needed in order to cleanup kernel for FOS logic in 7.0.
Renamed the class a bit while at it from `..MVC\Symfony\Security\User\ContextProvider\RoleContextProvider` to `ContextProvider\RoleIdentify`.

Maybe `RoleIdentity` might be better tough.

Todo:
- [x] Will have to drop extending kernel `PurgeClientInterface` here if we want to remove that from kernel


andrerom:
So seems we need to setup functional tests on this repo.

from @emodric:
> I now get `The service `fos_http_cache.user_context.hash_generator"` has a dependency on a non-existent service `ezpublish.user.identity_definer.role_id` after updating http cache repo to `0843f51`


Input welcome _(I'm about to leave for airport so a bit handicapped atm, alt revert this and we redo it)_

andrerom:
Reverted

PR #46, opened 2017-12-27
Move UserContextSubscriber logic to ezplatform-http-cache

andrerom:
Another thing needed for https://github.com/ezsystems/ezpublish-kernel/pull/2194

Look into if we could already reuse more of FosHttpCache here _(for instance it's getDefaultSubscribers Purge & Refresh Subscriber, as commented on inline here now)_, however that won't be possible before we move to FosHttpCache 2.x.

### DIFF
--- a/src/Proxy/UserContextSubscriber.php
+++ b/src/Proxy/UserContextSubscriber.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
+ * File containing the UserContextSubscriber class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\Proxy;
+
+use FOS\HttpCache\SymfonyCache\UserContextSubscriber as BaseUserContextSubscriber;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Extends UserContextSubscriber from FOSHttpCache to include original request.
+ *
+ * {@inheritdoc}
+ */
+class UserContextSubscriber extends BaseUserContextSubscriber
+{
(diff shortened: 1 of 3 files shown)


src/Proxy/UserContextSubscriber.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * {@inheritdoc}
+ */
+class UserContextSubscriber extends BaseUserContextSubscriber
+{
+    protected function cleanupHashLookupRequest(Request $hashLookupRequest, Request $originalRequest)
+    {
+        parent::cleanupHashLookupRequest($hashLookupRequest, $originalRequest);
+        // Embed the original request as we need it to match the SiteAccess.
+        $hashLookupRequest->attributes->set('_ez_original_request', $originalRequest);
```
andrerom: Is this still relevant @bdunogier / @lolautruche ? Ref:
https://github.com/ezsystems/ezpublish-kernel/blob/7.0/eZ/Publish/Core/MVC/Symfony/EventListener/SiteAccessMatchListener.php#L71-L74

andrerom:
Tests green, failure caused by packagist issue on missing stable packages for php-diff (needed by phpspec).

src/Proxy/UserContextSubscriber.php
bdunogier: It is indeed very contradictory. I don't have a good answer now. The only thing left to do is I think annotate the files, and see what was added when, and for what reason. Unless we're very unlucky, we should be able to make sense out of it :)

lolautruche: Yes it's still relevant AFAICS. The problem is probably the obsolete comment.
Comes from https://github.com/ezsystems/ezpublish-kernel/commit/880dc52d3c680f4e6ff5cdd1e559babda48a264b to fix https://jira.ez.no/browse/EZP-23632

PR #47, opened 2018-01-11
Fix: AnonymousRequestMatcher doesn't check session name prefix

blankse:
Requires and Description: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/434

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "ezsystems/ezpublish-kernel": "^6.13.2@dev || ^7.1@dev",
-        "friendsofsymfony/http-cache-bundle": "~1.2|^1.3.8",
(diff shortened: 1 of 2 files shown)


blankse:
@andrerom I rebased to master to avoid the conflict in the composer.json

andrerom:
Thanks!  And this now requires friendsofsymfony/http-cache-bundle 1.3.13 right?
And will it break without it? (asking as I'm thinking this should be applied to 0.4 also)

blankse:
Yes will break. Because the YML setting doesn't exist before v1.3.13

andrerom:
Tagged: https://github.com/ezsystems/ezplatform-http-cache/releases/tag/v0.5.0-rc4

Thanks @blankse !! :)

PR #53, opened 2018-02-19
EZP-28760: Change to use space seperated header

andrerom:
> [EZP-28760](https://jira.ez.no/browse/EZP-28760) Varnish xkey header using wrong separator

Header seems to be transformed to comma separated values per [rfc2616](https://tools.ietf.org/html/rfc2616#section-4.2)
by potentially Apache &/or others _(not seen this happen on nginx)_.

As Varnish xkey now has doc stating space separated header is supported,
change to rather use that to stay on the safe side of this.

### DIFF
--- a/src/Proxy/TagAwareStore.php
+++ b/src/Proxy/TagAwareStore.php
@@ -73,8 +73,12 @@ class TagAwareStore extends Store implements RequestAwarePurger
         $digest = $response->headers->get('X-Content-Digest');
         $tags = $response->headers->get('xkey', null, false);
 
-        // Handle both array based and string based header (her gotten as single item array with space separated string)
-        foreach (array_unique(count($tags) === 1 ? explode(' ', $tags[0]) : $tags) as $tag) {
+        if (count($tags) === 1) {
+            // Handle string based header (her gotten as single item array with space separated string)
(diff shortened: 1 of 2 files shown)


src/Proxy/TagAwareStore.php
```diff
@@ -69,11 +69,12 @@ public function write(Request $request, Response $response)
...
         $key = parent::write($request, $response);
 
-        // Now save tags
+        // Get tags in order to save them
         $digest = $response->headers->get('X-Content-Digest');
         $tags = $response->headers->get('xkey', null, false);
 
-        foreach (array_unique($tags) as $tag) {
+        // Handle both array based and string based header (her gotten as single item array with space separated string)
+        foreach (array_unique(count($tags) === 1 ? explode(' ', $tags[0]) : $tags) as $tag) {
```
adamwojs: IMHO more readable is simple `if` statement before loop:

```php
$tags = $response->headers->get('xkey', null, false);
if (count($tags) === 1) {
    // Handle string based header (her gotten as single item array with space separated string)
    $tags = explode(' ', $tags[0]);
}

foreach (array_unique($tags) as $tag) {
    //...
}
```

PR #55, opened 2018-02-26
EZP-28580: HTTP Cache purge when URLService\UpdateURLSignal is emitted

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28585
> Requires: https://github.com/ezsystems/ezpublish-kernel/pull/2259

## Description 

Implementation of HTTP cache purge when `URLService\UpdateURLSignal` is emitted. Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2259

## TODO: 
- [x] Raise the `ezpublish-kernel` requirement when  https://github.com/ezsystems/ezpublish-kernel/pull/2259 will be merged

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "~6.7.7@dev || ^6.12.1@dev || ^7.0@dev",
+        "ezsystems/ezpublish-kernel": "^6.13.2@dev || ^7.1@dev",
         "friendsofsymfony/http-cache-bundle": "~1.2|^1.3.8",
         "symfony/symfony": "^2.7 | ^3.1"
     },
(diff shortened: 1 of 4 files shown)


andrerom:
You don't need to raise kernel requirement, can do like in RemoveTranslationSlotTest

adamwojs:
@andrerom  I agree in case of test, but how about slot implementation which depends on `urlChanged` property introduced in https://github.com/ezsystems/ezpublish-kernel/pull/2259?

andrerom:
@adamwojs If we can manage to do feature detection there that would be great, until now we have bumped kernel requirement every time, but I like the approach here better.

But indeed, property additions are slightly different, but in this case we can do indexing if property is not set, and if set and true, newer kerne version = more optimal operation.

But if you prefer kernel bump we can also do that.

andrerom:
+1 (I fixed CS after merge)

PR #56, opened 2018-02-26
EZP-28872: Clear cache for non main locations on state/section changes

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28872


vidarl:
Approved, but tests needs to be fixed ....

PR #51, opened 2018-02-18
EZP-28848: Enable using FOSHttpCache for tagging responses

andrerom:
Enable usage of [fos_httpcache_tag()](http://foshttpcachebundle.readthedocs.io/en/1.3/features/tagging.html#tagging-from-twig-templates) and FOS TagHandler in order to be able to solve tagging relations which is template logic, added doc on how.

Todo:
- [x] Manually test usage of fos_httpcache_tag()

Followup:
- PR on overloading template output to tag responses _(as kernel is meant to know nothing of http cache)_

PR #58, opened 2018-03-19
EZP-28970: Use addTags() to avoid serializing tags several times

andrerom:
Since #51  it has been possible to use FOSHttpCache TagHandler, however internally in ezplatform-http-cache we don't take advantage of this and are serializing tags back and forward from response headers (string value) on every tag added.

So some small cleanup to:
- Use addTags()
- Make sure tag handler instance is the same so tags are correctly added to the instance being used to write headers

Also adjust order of response events, reason for this is that FOSHttpCacheBundle writes tags on subscriber with priority 0, so we need to make sure we add tags before then.

Also also; fix kernel requirement as v0.4 will not work on 7.1 and up _(v0.5 is for that)_.

Todo:
- [x] fix tests
- [x] Remaining failure seems to be related to kernel changes on REST _(checking)_


src/Handler/TagHandlerInterface.php
```diff
@@ -6,5 +6,20 @@
...
+     * @deprecated Use addTags()
+     */
     public function addTagHeaders(Response $response, array $tags);
+
+    /**
+     * @param array $tags
+     *
+     * @return $this
+     */
+    public function addTags(array $tags);
```
bdunogier: Adding methods to interfaces is a BC break (existing implementations of this interface will break on upgrade).

We could cheat, ignore `$response` in the implementation, and just call `parent::addTags()`. 
Ideally, we would add another interface, that has the correct method, and have it implemented the current one. But what would we call it ? The good name is already taken :)

andrerom: ResponseTaggerInterface, reflecting naming in FOS 2.x:
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ResponseTagger.php

Current one reflect naming in 1.4:
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/1.4/src/Handler/TagHandler.php

vidarl:
Look okay to me, but cannot say if it will brake anyting or not without testing

But looks like it will defently breake fastly unless that one is changed accordingly:
https://github.com/ezsystems/ezplatform-http-cache-fastly/search?utf8=%E2%9C%93&q=TagHandlerInterface&type=

andrerom:
> But looks like it will definitely break Fastly unless that one is changed accordingly:
https://github.com/ezsystems/ezplatform-http-cache-fastly/search?utf8=%E2%9C%93&q=TagHandlerInterface&type=


@vidarl indeed, we can hold this PR of for 1.13.2

vidarl:
But since it is a BC, maybe only a subject for 0.5 ( ezplatform 2.1 )? Give that we get the fix in fastly in time ... ?

emodric:
@andrerom Does #51 mean that we can use FOS tagger on both 0.4/7.0 and 0.5/7.1?

andrerom:
@emodric Yes. But 0.5 will also work on 6.13 and 7.0 btw.

emodric:
:+1: Good to know!

andrerom:
@emodric As stated in `docs/using_tags.md` TagsHandler class / namespace name will change once we upgrade to FOSHttpCache 2.x.

Could you share where you intend to use it? Is twig operator or X-Location-Id an option for the use case?

emodric:
I don't have issues when it changes, we will adopt our code :)

The usecase is tagging an ESI block that renders the main menu, once the eZ object responsible for the menu is republished. I migrated that code away from `X-Location-Id` couple of months ago becuase of the deprecation.

PR #60, opened 2018-03-22
[0.4] Ajust composer and travis prior to 2.1 release

andrerom:
0.4 does not work on 2.1 as the changes for that is part of 0.5, so
adjusts both composer and travis to reflect that.

Side: This might fail for unrelated reason on PHP 5.6 with 1.13, if
so then we know for sure #58 is not causing it's failues.


andrerom:
Failing where we expect it to, merging this to make sure people upgrading to 2.1 does not get into problems with using 0.4

PR #61, opened 2018-03-22
[Varnish] Use PURGE for purgeAll()

andrerom:
Found when reviewing code for another issue atm, now that purgeAll() is
supposed to work by invalidating a tag, it should use PURGE _(using
invalidatePath() like code above)_, and not BAN _(as used by invalidate())_.

### DIFF
--- a/src/PurgeClient/VarnishPurgeClient.php
+++ b/src/PurgeClient/VarnishPurgeClient.php
@@ -50,6 +50,9 @@ class VarnishPurgeClient implements PurgeClientInterface
 
     public function purgeAll()
     {
-        $this->cacheManager->invalidate(['key' => 'ez-all']);
+        $this->cacheManager->invalidatePath(
+            '/',
+            ['key' => 'ez-all', 'Host' => empty($_SERVER['SERVER_NAME']) ? 'localhost' : $_SERVER['SERVER_NAME']]
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


vidarl:
It breaks the tests...

Besides that, I think that function is never used, besides in the test.
So, do we need it?

vidarl:
Thinking about it, I think we should indeed keep the call API wise.

Background for the call. AFAIK, the call was used a while back when a cache:clear would also clear varnish cache ( that was done doing via purgeAll() I believe).
Unfortunately, I don't remember when we stopped doing that

andrerom:
> Unfortunately, I don't remember when we stopped doing that

Could be we stopped doing that with 2.1 as the change on removing FOS http cache code from kernel removed InstantCachePurger: https://github.com/ezsystems/ezpublish-kernel/pull/2194/files#diff-b1786c51bff3987a359c019bce7fd6f0

If we should do this or not I'm a but torn on.


As for this method, I would say we keep it for now, as Faslty have seperate end point to invalidate everything right?

But besides that, with all latests changes users should be able to clear everything using:
`bin/console fos:httpcache:invalidate:tag ez-all`

andrerom:
@vidarl Tests fixes, the remaining failures _(behat on PHP 5.6)_ is not relevant.

PR #62, opened 2018-03-22
EZP-28993: Varnish cache is not updated after modifying Content

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZP-28993

## Description

This PR backports `HttpCachePass` removed in https://github.com/ezsystems/ezpublish-kernel/pull/2194/files#diff-469a42b65cf73f41f143b52df89d1e70

### DIFF
--- a/src/DependencyInjection/Compiler/HttpCachePass.php
+++ b/src/DependencyInjection/Compiler/HttpCachePass.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * File containing the HttpCachePass class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * HttpCache related compiler pass.
+ *
+ * Ensures Varnish proxy client is correctly configured.
+ */
+class HttpCachePass implements CompilerPassInterface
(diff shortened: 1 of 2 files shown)


kmadejski:
Can't ask for the review, so @andrerom @vidarl take a look, please.

src/DependencyInjection/Compiler/HttpCachePass.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private function processCacheManager(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('ezplatform.http_cache.cache_manager')) {
+            return;
+        }
+
+        // Remove so 7.0 & 6.x kernels skip HttpCachePass
+        $container->removeAlias('ezpublish.http_cache.cache_manager');
```
vidarl: You are 100% sure this compiler pass is executed before the one in kernel, right?

vidarl: ( I am 99% sure it is correct, as EzSystemsPlatformHttpCacheBundle is included before eZ\Bundle\EzPublishCoreBundle\EzPublishCoreBundle() in AppKernel)

kmadejski: From the logical point of view, it should be ordered according to the bundles order in AppKernel. I can't find exact information about this behavior in the Symfony docs, but perhaps this description shed some light on this (ref: https://symfony.com/blog/new-in-symfony-3-2-compiler-passes-improvements):

> However, you can't select the execution order of the compiler passes in each step, so they are executed in the same order as they are found by Symfony. In some cases this is not acceptable, so Symfony 3.2 allows you to set the priority of the compiler passes.

Based on that, I'm assuming that if we don't set exact priority, it will be executed in accordance with the order.

Plopix:
Don't think it is working. I have just tested, and Varnish is not purged.

andrerom:
@Plopix We unfortunately need more info then that, see https://jira.ez.no/browse/EZP-28993

PR #64, opened 2018-04-20
Fix user hash tag being applied to main response when using Sf Proxy

andrerom:


src/EventSubscriber/UserContextSubscriber.php
```diff
@@ -48,9 +47,8 @@ public function tagUserContext(FilterResponseEvent $event)
             return;
         }
 
-        // Improvement potential (mainly relevant if we plan to increase hash ttl notably):
-        // - Get user id to also add tag by id, so user (un)assign don't need to clear all hashes
-        // - If done in layer generating this, we could also tag role id so changes to roles only expire affected hashes
-        $this->tagHandler->addTags(['ez-user-context-hash']);
+        // We need to set tag directly on repsonse here to make sure this does not also get applied to the main request
+        // when using Symfony Proxy, as tag hander does not clear tags between requests.
```
kmadejski: Nitpick: `hander` -> `handler`

PR #68, opened 2018-06-14
[Travis] Run REST functional tests on PR's

andrerom:


andrerom:
Commits merge to master _(fast forward, so no explicit merge commit)_.

PR #69, opened 2018-06-14
EZP-29297: 404 error pages for hidden locations should be tagged

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-29297](https://jira.ez.no/browse/EZP-29297)
| **Bug**| yes
| **New feature**    | no
| **Target version** | master (`0.7`)
| **BC breaks**      | no
| **Tests pass**     | yes/no
| **Doc needed**     | no

These changes are related to: https://github.com/ezsystems/ezpublish-kernel/pull/2350

### DIFF
--- a/src/EventSubscriber/HiddenLocationExceptionSubscriber.php
+++ b/src/EventSubscriber/HiddenLocationExceptionSubscriber.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\EventSubscriber;
+
+use eZ\Publish\Core\MVC\Exception\HiddenLocationException;
+use EzSystems\PlatformHttpCacheBundle\ResponseConfigurator\ResponseCacheConfigurator;
+use EzSystems\PlatformHttpCacheBundle\ResponseTagger\Value\LocationTagger;
+use Symfony\Bundle\TwigBundle\Controller\ExceptionController;
+use Symfony\Component\Debug\Exception\FlattenException;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class HiddenLocationExceptionSubscriber implements EventSubscriberInterface
+{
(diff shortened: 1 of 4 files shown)


src/EventSubscriber/HiddenLocationExceptionSubscriber.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        if (!$event->getException() instanceof HiddenLocationException) {
+            return;
+        }
+
+        /** @var HiddenLocationException $exception */
+        $exception = $event->getException();
+
+        $response = $this->exceptionController->showAction($event->getRequest(), FlattenException::create($exception));
+        $this->responseTagger->tag($this->responseConfigurator, $response, $exception->getLocation());
```
andrerom: please rebase thins on top of #67 so you don't need to create response

andrerom: As in this will also avoid you having to inject _twig.controller.exception_ right?

kmadejski: Both correct. Thanks to https://github.com/ezsystems/ezplatform-http-cache/pull/67 the solution has been significantly simplified.

kmadejski:
@andrerom  / @adamwojs / @alongosz tests here didn't pass due to dependency on the kernel. This comes from the new spec which I added for this solution. I think we can safely skip forcing composer to load correct kernel branch here and act as if all tests passed 😉.

alongosz:
Control question: does the feature require new kernel or just the tests? 

If it's the former one then [this line](https://github.com/ezsystems/ezplatform-http-cache/blob/master/composer.json#L13) needs to be changed to `"ezsystems/ezpublish-kernel": "^7.2@dev"`, which is probably not what you want. Just asking... :)

kmadejski:
@alongosz this https://github.com/ezsystems/ezplatform-http-cache/pull/69/files#diff-27a7b10b0921b701f0f9be7a6081bcecR18 spec depends on `HiddenLocationException` introduced in: https://github.com/ezsystems/ezpublish-kernel/pull/2350/files#diff-39129cd096d79d4468524079d3dc6fcaR12

If we really want to make this tests pass, then I would need to change the line which you've mentioned to something like `"dev-ezp_29297 as 6.13.2"`. Do you want me to change it to confirm?

andrerom:
@kmadejski can you try to skip this if class is missing?
https://github.com/phpspec/phpspec/pull/296 _(try to do it in let() perhaps)_

PR #72, opened 2018-11-28
EZP-29817: Varnish - Purge requests with tokens

damianz5:
Meta repo changes:

default_parameters.yml
```
    varnish_invalidate_token: '%env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN)%'

    env(HTTPCACHE_VARNISH_INVALIDATE_TOKEN): ~
```

ezplatform.yml
```
ezpublish:
...
	system:
	...
		site_group: (same for admin_group)
            		http_cache:
			        purge_servers: ['%purge_server%']
+             			varnish_invalidate_token: '%varnish_invalidate_token%'
```

.env
```
# Enable varnish purging invalidate token by setting valid token
#HTTPCACHE_VARNISH_INVALIDATE_TOKEN=""
```

https://jira.ez.no/browse/EZP-29817

### DIFF
--- a/docs/varnish/vcl/varnish4.vcl
+++ b/docs/varnish/vcl/varnish4.vcl
@@ -135,9 +135,7 @@ sub ez_purge {
 
     # Support how purging was done in earlier versions, this is deprecated and here just for BC for code still using it
     if (req.method == "BAN") {
-        if (!client.ip ~ invalidators) {
-            return (synth(405, "Method not allowed"));
-        }
+        call ez_purge_acl;
 
         if (req.http.X-Location-Id) {
             ban("obj.http.X-Location-Id ~ " + req.http.X-Location-Id);
@@ -149,9 +147,7 @@ sub ez_purge {
     }
 
     if (req.method == "PURGE") {
-        if (!client.ip ~ invalidators) {
-            return (synth(405, "Method not allowed"));
-        }
+        call ez_purge_acl;
 
         # If http header "key" is set, we assume purge is on key and you have Varnish xkey installed
         if (req.http.key) {
@@ -164,6 +160,19 @@ sub ez_purge {
 
         # if not, then this is a normal purge by url
         return (purge);
+    }
+}
+
+sub ez_purge_acl {
+    if (req.http.x-purge-token) {
+        # Use the line above instead of  line with 'std.getenv' if your Varnish version is lower than 6.2 and doesn't support std.getenv,
+        # change the token "_YOUR_TOKEN_HERE_" before enabling it.
+        # if (req.http.x-purge-token != "_YOUR_TOKEN_HERE_") {
+        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
+            return (synth(405, "Method not allowed"));
+        }
+    } else if  (!client.ip ~ invalidators) {
+        return (synth(405, "Method not allowed"));
     }
 }
 
--- a/docs/varnish/vcl/varnish4.vcl
+++ b/docs/varnish/vcl/varnish4.vcl
@@ -164,17 +164,33 @@ sub ez_purge {
 }
 
 sub ez_purge_acl {
-    if (req.http.x-purge-token) {
-        # Use the line above instead of  line with 'std.getenv' if your Varnish version is lower than 6.2 and doesn't support std.getenv,
-        # change the token "_YOUR_TOKEN_HERE_" before enabling it.
-        # if (req.http.x-purge-token != "_YOUR_TOKEN_HERE_") {
-        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
-            return (synth(405, "Method not allowed"));
-        }
-    } else if  (!client.ip ~ invalidators) {
+    if  (!client.ip ~ invalidators) {
         return (synth(405, "Method not allowed"));
     }
 }
+
+## Alternative 1: ez_purge_acl for purging by Token (Varnish >= 5.1, std.getenv support)
+#sub ez_purge_acl {
+#    if (req.http.x-purge-token) {
+#        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
+#            return (synth(405, "Method not allowed"));
+#        }
+#    } else if  (!client.ip ~ invalidators) {
+#        return (synth(405, "Method not allowed"));
+#    }
+#}
+
+## Alternative 2: ez_purge_acl for purging by Token (Varnish <= 5.1, token value in vcl)
+#sub ez_purge_acl {
+#    if (req.http.x-purge-token) {
+#        # change the token "_YOUR_TOKEN_HERE_" before enabling it.
+#        if (req.http.x-purge-token != "_YOUR_TOKEN_HERE_") {
+#            return (synth(405, "Method not allowed"));
+#        }
+#    } else if  (!client.ip ~ invalidators) {
+#        return (synth(405, "Method not allowed"));
+#    }
+#}
 
 // Sub-routine to get client user context hash, used to for being able to vary page cache on user rights.
 sub ez_user_context_hash {
--- a/docs/varnish/vcl/varnish4.vcl
+++ b/docs/varnish/vcl/varnish4.vcl
@@ -164,33 +164,16 @@ sub ez_purge {
 }
 
 sub ez_purge_acl {
+//    if (req.http.x-purge-token) {
+//        #  Won't work on Varnish <= 5.1, if needed you can hardcode a secret token here instead of std.getenv() usage
+//        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
+//            return (synth(405, "Method not allowed"));
+//        }
+//    } else if  (!client.ip ~ invalidators) {
     if  (!client.ip ~ invalidators) {
         return (synth(405, "Method not allowed"));
     }
 }
-
-## Alternative 1: ez_purge_acl for purging by Token (Varnish >= 5.1, std.getenv support)
-#sub ez_purge_acl {
-#    if (req.http.x-purge-token) {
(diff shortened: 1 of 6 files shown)


andrerom:
Cool, didn't know it was planned.

As for VCL, not same approach, but started looking into it today and have this if you need some pointers on one possible approach: https://gist.github.com/andrerom/c1f123fe65b577ca915c57fa6ceaf835

src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
```diff
@@ -33,6 +33,14 @@ public function addSemanticConfig(NodeBuilder $nodeBuilder)
                         ->example(array('http://localhost/', 'http://another.server/'))
                         ->requiresAtLeastOneElement()
                         ->prototype('scalar')->end()
+                    ->end()
+                    ->scalarNode('purge_auth_header')
+                        ->info('Header name for authenticated purge')
```
andrerom: maybe specify this is for varnish specifically? Same below

src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
```diff
@@ -33,6 +33,14 @@ public function addSemanticConfig(NodeBuilder $nodeBuilder)
...
+                        ->defaultValue('X-PURGE-AUTH')
+                    ->end()
+                    ->scalarNode('purge_auth_key')
+                        ->info('Purge authentication key')
```
andrerom: Besides mentioning it is for Varnish, also make it clear it's optional as normally auth is done using ACL (ip based).

andrerom:
Once we have updated VCL here, we can get @Plopix, or one of us to update https://github.com/ezsystems/ezplatform/pull/315/files as well to finish this code wise (doc will also be needed).

damianz5:
@andrerom It was PoC at the beginning, but I will remove this label - review is required, I will continue to work on this regarding the reviews

andrerom:
Ok, I'd like changed so VCL so we can see the full picture as part of the review, feel free to use from example gist link above.

damianz5:
Note: [platform.sh] check if ACL_INVALIDATE_TOKEN (or any env in general) will be injected to varnish node, and not only to app node.

docs/varnish/vcl/varnish4.vcl
```diff
@@ -13,6 +13,7 @@
 vcl 4.0;
 import std;
 import xkey;
+import var;
```
andrerom: can skip this as long as we use env

Plopix:
Regargind this: https://github.com/ezsystems/ezplatform/pull/315/files

I think they will or they have changed something else regarding the relationship.
I say that based on a conversation I had with @crell but I am unsure of the status if they kept it or not.

Also with this token I don't think we need a relationship as the whole point is to remove it. (to avoid the circular reference)

I think we need Platform.sh in the loop to confirm everything.

Then I will be happy to update the PR!

But that's cool, I can't wait to have varnish ready

andrerom:
You are bringing up a good point here @plopix, we/Symfony rely on IP to denote which Proxy we trust. This implies still need for circular relationship unless there are suggestions on how we can avoid it.

One way is to add support for host names in app.php for trusted proxies (look them up). But if we can avoid that, it’d be great.

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -7,20 +7,30 @@
...
 use FOS\HttpCacheBundle\CacheManager;
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
 
 /**
  * Purge client based on FOSHttpCacheBundle.
  */
 class VarnishPurgeClient implements PurgeClientInterface
 {
+    public const INVALIDATE_TOKEN_NAME_PARAM = 'http_cache.varnish_invalidate_token_name';
+    public const INVALIDATE_TOKEN_PARAM = 'http_cache.varnish_invalidate_token';
```
mnocon: Side note (maybe you're aware of this already) : this bundle is also used with eZ Platform 1.13, which means PHP 5.6: access modifiers for const can't be used 😛

docs/varnish/vcl/varnish4.vcl
```diff
@@ -167,6 +163,35 @@ sub ez_purge {
...
+#sub ez_purge_acl {
+#    if (req.http.x-purge-token) {
+#        # change the token "_YOUR_TOKEN_HERE_" before enabling it.
+#        if (req.http.x-purge-token != "_YOUR_TOKEN_HERE_") {
+#            return (synth(405, "Method not allowed"));
+#        }
+#    } else if  (!client.ip ~ invalidators) {
+#        return (synth(405, "Method not allowed"));
+#    }
+#}
```
andrerom: In retrospect this could have just been:
```
sub ez_purge_acl {
//    if (req.http.x-purge-token) {
//        #  Won't work on Varnish 4.1, if needed you can hardcode a secret token here instead of std.getenv() usage
//        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
//            return (synth(405, "Method not allowed"));
//        }
//    } else if  (!client.ip ~ invalidators) {
    if  (!client.ip ~ invalidators) {
        return (synth(405, "Method not allowed"));
    }
}
```
With corresponding simplification in parameters.vcl.

damianz5: Done

docs/varnish/vcl/varnish4.vcl
```diff
@@ -167,6 +163,18 @@ sub ez_purge {
     }
 }
 
+sub ez_purge_acl {
+//    if (req.http.x-purge-token) {
```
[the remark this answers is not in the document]
damianz5: `HTTPCACHE_VARNISH_INVALIDATE_TOKEN_NAME` has default value https://github.com/ezsystems/ezplatform-http-cache/pull/72/files#diff-bf4ea0152820eeb715fdd53160bca64dR39 and can be ommited in meta repo config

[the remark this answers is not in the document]
vidarl: Yes, but if the header names used is literally always going to `x-purge-token`, are there any need for it to be exported as config parameter and env variable?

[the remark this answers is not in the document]
damianz5: I think it should be possible to change the token name, but no need to set the same value as default one, I've modified the description for meta repo.

docs/varnish/vcl/varnish4.vcl
```diff
@@ -167,6 +163,18 @@ sub ez_purge {
...
+//        #  Won't work on Varnish <= 5.1, if needed in 4.1 you can hardcode a secret token here instead of std.getenv() usage
+//        if (req.http.x-purge-token != std.getenv("ACL_INVALIDATE_TOKEN")) {
```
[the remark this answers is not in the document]
damianz5: changed 👍

docs/varnish/vcl/varnish4.vcl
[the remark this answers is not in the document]
andrerom: I think I agree with @vidarl here, don't see a need for allowing to set different header name.
We don't expose setting to use different method for PURGE for instance.

Also be aware this might be thrown away in the next half year when we get forced to move to FosHttpCache 2 support (for Symfony4), which means lots of logic in this repo should be dropped.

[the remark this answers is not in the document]
damianz5: done

damianz5:
@andrerom @vidarl final review please

docs/varnish/vcl/varnish4.vcl
```diff
@@ -167,6 +163,18 @@ sub ez_purge {
     }
 }
 
+sub ez_purge_acl {
+//    if (req.http.x-purge-token) {
+//        #  Won't work on Varnish <= 5.1, if needed in 4.1 you can hardcode a secret token here instead of std.getenv() usage
+//        if (req.http.x-purge-token != std.getenv("HTTPCACHE_VARNISH_INVALIDATE_TOKEN")) {
```
vidarl: I have not tested, but this looks like a security hole

If server is configured to use whitelist (invalidators), and evil user sends x-purge-token header with empty value, this `if` might match if env variable `HTTPCACHE_VARNISH_INVALIDATE_TOKEN` is not set?

damianz5: I think if `x-purge-token` will be empty - the if condition `if (req.http.x-purge-token) {` will return false and second condition (`req.http.x-purge-token != std.ge...`) will not be called

andrerom: This is why it's not enabled by default.
As enabling it means you should always inject env.

But, as p.sh does not support injecting it by env, which is where we need this. We can change example here to just have hardcoded token to avoid anyone stumbling into such an issue by mistake.

Avoid having to document any differences between 4.1 and higher as well.

andrerom: Or, what was the outcome so far of discussions with p.sh?

vidarl: No reply from ph.sh yet. I have created a ticket in their jira, requesting us to be able to define env variables in the varnish container

vidarl: I have now tested sending token header with empty value, and indeed it seems to work as we intended.

So, since that is the case, why comment out the lines in the first place...?

andrerom: See discussion above, it's because std.getenv is not defined on Varnish 4.1 (LTS).
We can drop support for 4.x in a future major, but not in a minor where we need this.

vidarl: acknowledged

andrerom:
Merging for followup on aligning VCL files. And progressing the work in meta repo once platform.sh adds more convenient ways to inject secrets to Varnish, ref: https://github.com/ezsystems/ezplatform/pull/346

PR #73, opened 2018-12-18
EZP-29194: Improve repo prefix handling for purgeAll() and user context hash tagging

andrerom:
Issue: https://jira.ez.no/browse/EZP-29194

Followup to fix the following edge cases:
* Prefix tag for user context hash, so it can be correctly purged when repo prefixes er in use
* Always tag responses with unprefixed `ez-all`, as this is used in `purgeAll()` on Varnish and Fastly. Makes sure `purgeAll()` works the same in all Proxies; clearing _all_ cache across all repositories.
  * Note: To clear all cache but only for a specific repo, use _(`purge(['ez-all'])`)_ or from command line using fos command to clear by `ez-all` tag, these calls will/should be prefixed.

Follow up on this is to add a optional cache clearer for symfony that takes advantage of the fixes in purgeAll() to call that when user is calling `bin/console cache:clear`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "extra": {
         "branch-alias": {
             "dev-master": "0.8.x-dev",
-            "dev-tmp_ci_branch": "0.7.x-dev"
+            "dev-tmp_ci_branch": "0.8.x-dev"
         }
     }
 }
--- a/spec/Handler/TagHandlerSpec.php
+++ b/spec/Handler/TagHandlerSpec.php
@@ -95,7 +95,7 @@ class TagHandlerSpec extends ObjectBehavior
 
     public function it_tags_all_tags_we_add_and_prefix_with_repo_id(Response $response, ResponseHeaderBag $responseHeaderBag)
     {
-        $responseHeaderBag->set('xkey', Argument::exact('intranet_ez-all intranet_location-4 intranet_content-4 intranet_path-2'))->shouldBeCalled();
+        $responseHeaderBag->set('xkey', Argument::exact('intranet_ez-all intranet_location-4 intranet_content-4 intranet_path-2 ez-all'))->shouldBeCalled();
 
         $this->addTags(['location-4', 'content-4']);
         $this->addTags(['path-2']);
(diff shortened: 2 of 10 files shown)


andrerom:
```
In ExceptionConversion.php line 104:
                  
  Database error  
                  
In AbstractMySQLDriver.php line 42:
                                                                               
  An exception occurred while executing 'SELECT `ezuser`.`contentobject_id`,   
  `ezuser`.`login`, `ezuser`.`email`, `ezuser`.`password_hash`, `ezuser`.`pas  
  sword_hash_type`, `ezuser_setting`.`is_enabled`, `ezuser_setting`.`max_logi  
  n` FROM `ezuser` LEFT JOIN `ezuser_setting` ON `ezuser_setting`.`user_id` =  
   `ezuser`.`contentobject_id` WHERE LOWER(`ezuser`.`login`) = :placeholder1'  
   with params ["admin"]:                                                      
                                                                               
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'ezp.ezuser' does  
  n't exist   
```

@adamwojs rings a bell?

andrerom:
@vidarl Passing and ready for your review _(behat failure is unrelated to PR)_

andrerom:
> So, prefix on default install will be "default_" ?

no, default have a empty prefix afaik.

But yes, even when this is not set we have had cases where customers have reached the limit on Varnish. So @kmadejski and/or @konradoboza added doc on how to extend the limit as it's configurable. Last followup on that here: https://github.com/ezsystems/developer-documentation/pull/491/files

Limit for Fastly seems to be 16Kb:
https://docs.fastly.com/guides/debugging/resource-limits#request-and-header-limits


So seems to be twice the [default limit of Varnish]( which is twice the limit of Varnish but), but as it's not configurable that might be an issue in extreme cases.

We have talked about maybe cutting down on the tag sizes and phase out the current tags in favour of shorter variants. Afaik that is the only thing we can do, besides documenting that people should use short repo names _(we can also log warnings about that here btw)_.

PR #74, opened 2019-01-16
Allow https scheme for ProxyClient

vidarl:
Our [default routes](https://github.com/ezsystems/ezplatform/blob/master/.platform/routes.yaml) on Platform.sh only have https. So in order to send purge requests on Platform.sh, we need to support sending purge requests via HTTPS

### DIFF
--- a/src/PurgeClient/VarnishPurgeClient.php
+++ b/src/PurgeClient/VarnishPurgeClient.php
@@ -53,7 +53,7 @@ class VarnishPurgeClient implements PurgeClientInterface
 
             $headers = [
                 'key' => $tag,
-                'Host' => empty($_SERVER['SERVER_NAME']) ? 'localhost' : $_SERVER['SERVER_NAME'],
+                'Host' => empty($_SERVER['SERVER_NAME']) ? parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST) : $_SERVER['SERVER_NAME'],
             ];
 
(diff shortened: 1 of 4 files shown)


src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -53,7 +53,7 @@ public function purge($tags)
 
             $headers = [
                 'key' => $tag,
-                'Host' => empty($_SERVER['SERVER_NAME']) ? 'localhost' : $_SERVER['SERVER_NAME'],
```
vidarl: `localhost` is not a good fallback. Won't work on Platform.sh ( and other setups to ) from CLI where purge server is not local host

vidarl:
@damianz5 : review ping

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -53,7 +53,7 @@ public function purge($tags)
...
+                'Host' => empty($_SERVER['SERVER_NAME']) ? parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST) : $_SERVER['SERVER_NAME'],
```
andrerom: Wondering if this should always be purge server??

andrerom:
Todo: Fix Unit test failures in this PR, if you want also deal with memory issue on PHP 5.6 ;)

vidarl:
Tests should pass now

PR #75, opened 2019-01-16
[TBD] Implement cache clearer to purge http cache on "cache:clear"

andrerom:
Adds support for purging all http cache when developer runs `cache:clear` command
_Takes advantage of the recent fixes on purgeAll()_

TBD: _Several people have asked for it, but I'm far from sure everyone wants it with no way to opt out._

### DIFF
--- a/src/CacheClearer/HttpCacheClearer.php
+++ b/src/CacheClearer/HttpCacheClearer.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\CacheClearer;
+
+use EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface;
+use Symfony\Component\HttpKernel\CacheClearer\CacheClearerInterface;
+
+/**
+ * Cache cleared for purging all http cache on cache:clear.
+ */
+class HttpCacheClearer implements CacheClearerInterface
+{
+    /**
+     * @var \EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface
+     */
+    protected $purgeClient;
+
+    public function __construct(PurgeClientInterface $purgeClient)
+    {
+        $this->purgeClient = $purgeClient;
+    }
+
+    public function clear($cacheDirectory)
+    {
+        // In the case of Varnish & Fastly this results in a expiry and not a purge
+        // Meaning the cache items might still be served up to the time of the grace/stale period
+
+        // @todo: Ideally should be a way to opt out of this when needed, but cache warmers is not a good match either
+        $this->purgeClient->purgeAll();
+    }
+}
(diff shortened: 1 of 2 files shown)


vidarl:
I think we should separate "code" cache,(like the symfony container, twig templates etc.) and content cache ( http cache, sessions and persistance cache (if stored in redis) etc.).

So I would prefer a second command which deletes content cache

vidarl:
@runelangseid: FYI

vidarl:
👎

danielmuehlbacher:
👎 No, I'd like a seperate command for all eZ content cache which is extendable like cache:clear and which clears http, persistence, ... any cache holding eZ content

andrerom:
@danielmuehlbacher @vidarl You are doing it wrong, add reaction on PR, not as comment ;)

andrerom:
> A separate command for each type of cache would be the best. It is then easy to combine them depending on the needs
> I guess it should be nice to have an option to clear all .. like app/console cache:clear --all or something like this :)

So this is what I intended with *1* extendable command with 👎  vote, it should allow to clear just one, or skip just one as well.

Examples:
- `explatform:[content-]cache:clear` => All "eZ Content cache"
- `explatform:[content-]cache:clear --http` => Only http
- `explatform:[content-]cache:clear --skip-http` => All but http

ilukac:
In that case I change my vote. I am not sure that default command should clear all caches

emodric:
I agree with @ilukac, under no circumstances should clearing persistence cache be a default action.

andrerom:
@ilukac I feel like you might misunderstand, I'm not talking about Symfony's `cache:clear`, I'm talking about a new one that can cover content cache needs.

If you have the need to clear a specific cache what I propose a bove will allow you to do that.
And there will always be possiblity to do cusotm stuff yourself, or call redis or Faslty directly to purge using bash if you for some reason want that, but the command above is to simplify use of among others this use case _and_ the once you mention:

As a developer, when I _(for some reason, including backup restore)_ have deployed database changes, I'd like to clear *all* content related changes

emodric:
In my opinion, default execution of such command should not be possible, since the command is dangerous. Instead, you would need to manually specify what you wish to clear, e.g:

`cache:clear --http`
`cache:clear --persistence`
`cache:clear --http --persistence`
`cache:clear --all`

This format would also allow adding other types of caches later on, like image alias caches and so on.

ilukac:
Got it, but I still think it is risky to have all content related caches purged by default. 
Persistence cache is super important in preventing big and numerous queries hitting db, and varnish is now an integral part of the app with no alternative. Clearing all content cache means exposing the db completely for a while.

andrerom:
@emodric I can agree with that.

> I think an option to specify a subtree is also needed
@florianalexandre Why is that? _(use case)_ That would be something completely different then cache type

florianalexandre:
In a multisite configuration, you might want to invalidate the cache of a specific site only.

andrerom:
> In a multisite configuration, you might want to invalidate the cache of a specific site only.

This opens up a few questions.. Should it by default use siteaccess defined for command _(default siteaccess is used unless `--siteaccess` is set)_, and should then also have a `--all-siteaccess` option?

PR #76, opened 2019-01-31
[Cloud feature] Purge token controller

vidarl:
This makes it possible for varnish to fetch the purge token from the backend. This is needed on Platform.sh as it is not possible to inject environment variables into varnish container.

With this change we don't need to use environment variables in .vcl so I have removed that part

### DIFF
--- a/docs/varnish/vcl/varnish4.vcl
+++ b/docs/varnish/vcl/varnish4.vcl
@@ -127,6 +127,8 @@ sub vcl_backend_response {
 // You may add FOSHttpCacheBundle tagging rules
 // See http://foshttpcache.readthedocs.org/en/latest/varnish-configuration.html#id4
 sub ez_purge {
+    // Retrieve purge token, needs to be here due to restart, match for PURGE method done within
+    call ez_invalidate_token;
 
     # Support how purging was done in earlier versions, this is deprecated and here just for BC for code still using it
     if (req.method == "BAN") {
@@ -159,13 +161,11 @@ sub ez_purge {
 }
 
 sub ez_purge_acl {
-//    if (req.http.x-purge-token) {
-//        #  Won't work on Varnish <= 5.1, if needed in 4.1 you can hardcode a secret token here instead of std.getenv() usage
-//        if (req.http.x-purge-token != std.getenv("HTTPCACHE_VARNISH_INVALIDATE_TOKEN")) {
-//            return (synth(405, "Method not allowed"));
-//        }
-//    } else if  (!client.ip ~ invalidators) {
-    if  (!client.ip ~ invalidators) {
+    if (req.http.x-invalidate-token) {
+        if (req.http.x-invalidate-token != req.http.x-backend-invalidate-token) {
+            return (synth(405, "Method not allowed"));
+        }
+    } else if  (!client.ip ~ invalidators) {
         return (synth(405, "Method not allowed"));
     }
 }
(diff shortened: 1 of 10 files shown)


vidarl:
Pending on https://github.com/ezsystems/ezplatform-http-cache/pull/74

docs/varnish/vcl/varnish4.vcl
```diff
@@ -224,7 +224,48 @@ sub ez_user_context_hash {
...
+            || req.http.x-backend-purge-token
+        )
+    ) {
+        return (synth(400));
+    }
+
+    if (req.restarts == 0 && req.method == "PURGE") {
+        set req.http.accept = "application/vnd.ezplatform.invalidate-token";
+
+        set req.url = "/http-cache/purgetoken";
```
andrerom: could perhaps be something ala `/_ez_http_purgetoken` to be more like fos approach.

docs/varnish/vcl/varnish4.vcl
```diff
@@ -224,7 +224,48 @@ sub ez_user_context_hash {
     }
 }
 
+// Sub-routine to get purge token.
+sub ez_invalidate_token {
+
```
andrerom: ws

src/EventListener/VaryHeaderListener.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+/**
+ * Class VaryHeaderListener
+ * Unfortunately, FOS\HttpCacheBundle\EventListener\UserContextSubscriber will set Vary header on all requests.
+ * This event listeners removes the $userIdentifierHeaders headers again in responses to any of the given $routes.
+ * For such routes, the controller should instead set the Vary header explicitly.
+ */
+class VaryHeaderListener implements EventSubscriberInterface
```
andrerom: `ConditionallyRemoveVaryHeaderListener` or similar

docs/varnish/vcl/varnish4.vcl
vidarl: ok, so going to use invalidate everywhere ( not a mix of purge and invalidate )..
So, let's use `_ez_http_invalidatetoken`

src/EventListener/VaryHeaderListener.php
vidarl: Fixed

vidarl:
Turns out that https://github.com/ezsystems/ezplatform-http-cache/pull/74 do not work properly on 1.13 as it has the kernel still injecting the fos' ProxyClient. Fixed in 369b56e4b4827d624b1645f8db8c847881f71bb7, but this commit is stricly not needed in 2.x. We might opt to branch of this bundle?

andrerom:
@vidarl Make sure the commit there works for both kernels, 0.8 will need that, for master/0.9 the commit can be reverted.

vidarl:
> @vidarl Make sure the commit there works for both kernels, 0.8 will need that, for master/0.9 the commit can be reverted.

Yeah. I have tested the code as-is now on both 1.13(ezplatform-demo) and 2.4(ezplatform-demo) and it works. Unable to test on master(ezplatform-demo) ATM because of issues with assets ( I guess someone will fix soon )

andrerom:
Ok, but seems you need to adjust specs for this change, @vidarl see failures.

vidarl:
Yeah, I already noticed .. I kinda fixed it the quick way by moving the code to HttpCachePass which already deals with ProxyClient somewhat

andrerom:
@vidarl So ready for merge or do you intend to do some more fixups on that?

vidarl:
I feel pretty done with it...

Plopix:
https://jira.ez.no/browse/EZP-30518 ping @andrerom @damianz5 @vidarl @SylvainGuittard ;-)

PR #77, opened 2019-02-22
EZP-30170: [VCL] Add own VCL for 2.5LTS to bump Varnish support there

andrerom:
Story: https://jira.ez.no/browse/EZP-30170
Needed for https://github.com/ezsystems/ezplatform/pull/349

### DIFF
--- a/docs/varnish/vcl/varnish5.vcl
+++ b/docs/varnish/vcl/varnish5.vcl
@@ -1,11 +1,11 @@
 // Varnish VCL for:
-// - Varnish 4.1 or higher (4.1LTS or 6.0LTS recommended, and is what we mainly test against)
+// - Varnish 5.1 or higher (6.0LTS recommended, and is what we mainly test against)
 //   - Varnish xkey vmod (via varnish-modules package, or via Varnish Plus)
-// - eZ Platform 1.13 with ezplatform-http-cache bundle
+// - eZ Platform 2.5LTS or higher with ezplatform-http-cache (this) bundle
 //
 // Make sure to at least adjust default parameters.yml, defaults there reflect our testing needs with docker.
 
-vcl 4.0;
+vcl 5.1;
 import std;
 import xkey;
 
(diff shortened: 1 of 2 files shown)


docs/varnish/vcl/varnish5.vcl
```diff
@@ -0,0 +1,269 @@
...
+        set req.http.X-Forwarded-Port = "443";
+    } else {
+        set req.http.X-Forwarded-Port = "80";
+    }
+
+    // Trigger cache purge if needed
+    call ez_purge;
+
+    // Don't cache requests other than GET and HEAD.
+    if (req.method != "GET" && req.method != "HEAD") {
```
alongosz: maybe `OPTIONS` also?

andrerom: [MSDN says no](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS), but afaik in our case we should safely be able to do that for limited time. Follow Up story? Might speed up REST :)

andrerom:
@mnocon At this point it would have been great if meta used VCL from here, so there where cross testing ;)

docs/varnish/vcl/varnish5.vcl
```diff
@@ -0,0 +1,269 @@
+// Varnish VCL for:
+// - Varnish 5.1 or higher (6.0LTS recommended, and is what we mainly test against)
+//   - Varnish xkey vmod (via varnish-modules package, or via Varnish Plus)
+// - eZ Platform 2.5LTS or higher with ezplatform-http-cache (this) bundle
+//
+// Make sure to at least adjust default parameters.yml, defaults there reflect our testing needs with docker.
+
+vcl 5.1;
```
mnocon: This value is not valid, according to docs it can be either `4.0` or `4.1`: https://varnish-cache.org/docs/6.1/whats-new/upgrading-6.0.html#vcl-4-0-and-4-1

Varnish fails to starts with this VCL.

andrerom: Fixed in a335b3d31e08b4b5a3ef051fd5075d0cfa97f0df

I set it to 4.1 which seems more future proof, if you see any issues with that shout out :)

andrerom:
Merging for now, to unblock meta PR.

@mnocon As discussed, could we have a follow up during certification so script in meta does install before building containers please so we can cover vcl and sorl config changes on travis behat runs in the future? /cc @lserwatka

PR #78, opened 2019-02-28
[Travis] Fix test failures

andrerom:
Aka, attempt to fix Travis failures.


andrerom:
Ok, I did not expect that to pass so easily.. But 🎉 
_(Change to use dist: xenial, cc @mnocon seems we can start to use it them ;))_

PR #80, opened 2019-03-14
Changed how FOS' proxyclient is overriden without breaking fastly bundle

vidarl:
Fastly bundle is already changing the fos_http_cache.proxy_client.varnish.class parameter in a [kernelpass](https://github.com/ezsystems/ezplatform-http-cache-fastly/blob/master/src/DependencyInjection/Compiler/KernelPass.php#L19). So when we in ezplatform-http-cache opted to replace the whole parameter with a new one, Fasty's proxyclient was never used


vidarl:
I need to retest 1.13, but I don's suspect any problems with this fix

vidarl:
@mnocon : FYI

vidarl:
Tested varnish on 1.13 and conformed that this still works

PR #81, opened 2019-03-28
EZP-30358: HTTP Cache purge when ContentService\{Hide,Reveal}ContentSignal is emitted

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-30358

## Description

Implementation of HTTP cache purge when `ContentService\HideContentSignal`/`ContentService\RevealContentSignal` is emitted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.3@dev",
+        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.5@dev",
         "friendsofsymfony/http-cache-bundle": "^1.3.13",
         "symfony/symfony": "^2.8.40 || ^3.4.11"
     },
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.5@dev",
+        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.3@dev",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.3@dev",
+        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.5@dev",
```
mnocon: Will it work for 2.4 branch? Ezplatform 2.4 uses both:
`"ezsystems/ezpublish-kernel": "~7.4.0@dev",`
`"ezsystems/ezplatform-http-cache": "~0.9.0@dev",`
(https://github.com/ezsystems/ezplatform/blob/2.4/composer.json#L36)
which will clash with this requirement if I'm not mistaken

lserwatka: Good catch Marek. This code can not crash 2.4 version. Is this doable?

andrerom: it will not afaik, if it does, then we also have same issue on 1.13

adamwojs: Actually, it should work fine without `composer.json` modification (signal is not emitted => slot will not be executed). We will check this during QA.

andrerom: Yeah, it's only solr having issues with this due to the use of abstract search slots, which is in retrospective not a good idea.

lserwatka:
@micszo Great news, can we merge it?

micszo:
@adamwojs did you want to fix CI before?

andrerom:
Travis failure is basically same 2.4 and 1.13 will experience so indeed we need to get that green :)

lserwatka:
Well done!

PR #82, opened 2019-05-06
EZP-30511: HttpCache purging failing if repository has been configured as "default"

andrerom:
JIRA: https://jira.ez.no/browse/EZP-30511

Adds check to see if repository is what is considered default, and also drops prefix in that case to avoid issues if user has configured repository as "default" in one siteaccess and forgot in another _(default value being null, which means the same as default repository)_.


src/RepositoryIdAwareTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle;
+
+/**
+ * Trait RepositoryIdAwareTrait.
+ *
+ * @internal For use in EzSystems\PlatformHttpCacheBundle package.
+ */
+trait RepositoryIdAwareTrait
```
adamwojs: IMHO we can introduce small utility service instead of using Traits here: 

```php
/**
 * @internal For use in EzSystems\PlatformHttpCacheBundle package.
 */
final class RepositoryPrefixProvider 
{	
	public function getRepositoryPrefix(string $repositoryId, array $repositories): string
	{
		// ...
		return empty($repositoryId) || $repositoryId === $default ? '' : $repositoryId . '_';
	}
}
```

andrerom: Indeed, I'll look into that when I better understand why the one REST test fails here.

andrerom: Pushed change for this, but opted to use config resolver as it might actually solve the failure _(if the failure is caused by scope change issues...)_.

andrerom:
@adamwojs Updated, and passing now with the changes ;)

PR #83, opened 2019-05-07
EZP-28034: Refactored to use EzPlatformRestBundle

ViniTou:
> JIRA: https://jira.ez.no/browse/EZP-28034

# Desc
Extract RestBundle and Common namespace to new package.

https://github.com/ezsystems/ezpublish-kernel/pull/2630
https://github.com/ezsystems/ezplatform-rest

# BC

as of `ezpublish-kernel 8.0`, support for php5.6 and 7.0 must be dropped. As it seems there is no way to include `ezplatform-rest` basing on `ezpublish-kernel` version. So this would have to be version `1.0`.


ViniTou:
Same story as in https://github.com/ezsystems/ezplatform-admin-ui/pull/989.
Circural dependencies between support-tools and admin-ui.

composer.json
```diff
@@ -37,7 +39,7 @@
         ]
     },
     "extra": {
-        "_ezplatform_branch_for_behat_tests": "master",
+        "_ezplatform_branch_for_behat_tests": "EZP-28032-rest-common",
         "branch-alias": {
             "dev-master": "0.10.x-dev",
```
alongosz: It also requires version bump, it's no longer technically possible to support multiple versions of ezpublish-kernel with these changes.
I guess `0.11.x-dev` should be fine // ping @andrerom

ViniTou: Yup, that's what I wrote in PR description but was opting for 1.0 to be somewhat consistent with other packages, but I have no strong opinion on that.

andrerom: 3.x version can aim to be 1.0 in the end if we'd like, but there are a tons of code to remove and adapt to use FOSHttpCache 2.x before we can make that stable.

ViniTou: Ok, so I will go with 0.11 for now

mnocon: I need to ask, out of curiosity - where will be 0.10 used? 2.5 uses 0.9, 0.10 was never released, do we really need 0.11?

ViniTou: Oh, well, I think we all didn't noticed that `0.10` was created for `3.0`

alongosz: > I need to ask, out of curiosity - where will be 0.10 used? 2.5 uses 0.9, 0.10 was never released, do we really need 0.11?

@mnocon :trophy:

lserwatka: We could go for 1.0.x-dev for 3.0, couldn't we?

alongosz: I think we should, but it would make most sense once we make it FOSHttpCache 2.x compatible, which we also planned to do anyway.

webhdx: We HAVE TO go for `1.0` with eZ Platform 3 since it won't be backwards compatible.

composer.json
```diff
@@ -10,7 +10,9 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.13.4@dev || ^7.3@dev || ^8.0@dev",
+        "ezsystems/ezpublish-kernel": "^8.0@dev",
+        "ezsystems/doctrine-dbal-schema": "^1.0@dev",
```
mnocon: I know that this is added because of:
```
If one of your dependencies has a dependency on an unstable package you need to explicitly require it as well, along with its sufficient stability flag.
```

I've encountered this issue when setting up dependencies for 3.0 - instead of adding the dependency everywhere we went with the approach of adding:
```
    "minimum-stability": "dev",
    "prefer-stable": true,
```
to the packages where it was needed (example: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/composer.json#L15). This change will have to be reverted after we release a stable version of `doctrine-dbal-schema`.

@alongosz @ViniTou maybe we should do the same here, to have an unified approach across all packages?

alongosz:
```
    "minimum-stability": "dev",
    "prefer-stable": true,
```
@mnocon hmm, that's interesting. I wasn't a huge fan of setting minimum stability to dev as it made composer install unstable versions of all packages, including 3rd party. So when combined with `prefer-stable` it's not the case?

tests/.travis/prepare_for_functional_tests.sh
```diff
@@ -19,5 +19,5 @@ if [ "$EZPLATFORM_BRANCH" = "1.13" ] ; then
     export RUN_INSTALL=1
     ./bin/.travis/trusty/setup_from_external_repo.sh $PACKAGE_BUILD_DIR "ezsystems/ezplatform-http-cache:dev-tmp_ci_branch"
 else
-    /bin/bash ./bin/.travis/trusty/setup_ezplatform.sh "${COMPOSE_FILE}" '' "${PACKAGE_BUILD_DIR}"
```
alongosz: AFAIR I've added it for a reason. Why do you need this change @ViniTou?

composer.json
mnocon: Yes, it installs dev only when there is no stable version available. It works in AdminUI, example build: https://travis-ci.org/ezsystems/ezplatform-admin-ui/jobs/536101037

You can see that it installs correct packages in @dev:
```
  - ...
  - Installing composer/ca-bundle (1.1.4): Loading from cache
  - Installing sensiolabs/security-checker (v5.0.3): Loading from cache
  - Installing sensio/distribution-bundle (v5.0.24): Loading from cache
  - Installing symfony-cmf/routing (2.1.0): Loading from cache
  - Installing ezsystems/ezpublish-kernel (dev-master 03e4495): Loading from cache
  - Installing ezsystems/repository-forms (dev-master 5ed5c97): Loading from cache
  - Installing ezsystems/ezplatform-richtext (dev-master eed18bb): Loading from cache
  - Installing symfony/polyfill-iconv (v1.11.0): Loading from cache
  - Installing egulias/email-validator (2.1.8): Loading from cache
  - Installing swiftmailer/swiftmailer (v6.2.1): Loading from cache
  - Installing symfony/swiftmailer-bundle (v3.2.6): Loading from cache
  - Installing ezsystems/ezplatform-user (dev-master 21d58df): Loading from cache
  - ...
```

Of course it works only because there are no stable versions of our packages for 3.0 - once we release them we will stop getting dev dependencies (that's why we will have to revert that after 3.0 stable is released). But I feel we would do the same with the requirement of `doctrine-dbal-schema` - it's not ideal to have it as a dependency in AdminUI/http-cache/other packages.

alongosz: > Yes, it installs dev only when there is no stable version available. It works in AdminUI, example build: https://travis-ci.org/ezsystems/ezplatform-admin-ui/jobs/536101037

I've also confirmed it just now. This is great workaround, @mnocon another :trophy: :grinning: 

@ViniTou so instead `"ezsystems/doctrine-dbal-schema": "^1.0@dev",` let's add after `require-dev` section: 
```json
    "minimum-stability": "dev",
    "prefer-stable": true,
```

composer.json
```diff
@@ -37,7 +39,7 @@
         ]
     },
     "extra": {
-        "_ezplatform_branch_for_behat_tests": "master",
+        "_ezplatform_branch_for_behat_tests": "EZP-28032-rest-common",
```
alongosz: Please go back to `master` here.

composer.json
andrerom: > We HAVE TO go for 1.0 with eZ Platform 3 since it won't be backwards compatible.

You don't have to _(0.x versions are all semantically major versions), but I prefer you go for 1.0.x-dev here for 3.x like @lserwatka prefer too. As it allows us bump to 0.10 and higher for 1.13/2.5 in case we need to introduce features/feature-changes to solve bugs, like for instance shorten tag length. .. :)

alongosz: Since it's already configured to be 3.0 compatible only, let's keep it as is and bump to 1.0 once we make FOSHttpCache 2.x compatibility.

PR #85, opened 2019-05-14
EZP-30554: Purge several keys at once to simplify debug & reduce round trips

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-30554](https://jira.ez.no/browse/EZP-30554)
| **Target version** | `0.9`
| **BC breaks**      | no
| **Doc needed**     | _maybe_

As of 2.5LTS we require Varnish 5.1 or higher, meaning we can take advantage of varnish-modules 0.10+ support for [several keys in header while purging](https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst#example). In order to:
- Simplify Varnish debugging when checking purge requests
- Reduce rountrips on between web server and varnish for purges

Given the package is also allowed to be used on 1.13, which supports Varnish 4.x. Should we ❓ 
- A. Expose a setting for this as currently is the case in this PR
- B. Avoid complexity of a temprary setting, and rather tell 1.13 + Varnish 4.x users to use ezplatform-http-cache 0.8.x.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/docs/varnish/vcl/varnish5.vcl
+++ b/docs/varnish/vcl/varnish5.vcl
@@ -1,6 +1,6 @@
 // Varnish VCL for:
 // - Varnish 5.1 or higher (6.0LTS recommended, and is what we mainly test against)
-//   - Varnish xkey vmod (via varnish-modules package 0,10.2 or higher, or via Varnish Plus)
+//   - Varnish xkey vmod (via varnish-modules package 0.10.2 or higher, or via Varnish Plus)
 // - eZ Platform 2.5LTS or higher with ezplatform-http-cache (this) bundle
 //
 // Make sure to at least adjust default parameters.yml, defaults there reflect our testing needs with docker.
--- a/docs/varnish/vcl/varnish5.vcl
+++ b/docs/varnish/vcl/varnish5.vcl
@@ -1,6 +1,6 @@
 // Varnish VCL for:
 // - Varnish 5.1 or higher (6.0LTS recommended, and is what we mainly test against)
-//   - Varnish xkey vmod (via varnish-modules package, or via Varnish Plus)
+//   - Varnish xkey vmod (via varnish-modules package 0,10.2 or higher, or via Varnish Plus)
 // - eZ Platform 2.5LTS or higher with ezplatform-http-cache (this) bundle
 //
 // Make sure to at least adjust default parameters.yml, defaults there reflect our testing needs with docker.
--- a/src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
+++ b/src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
@@ -37,10 +37,6 @@ class HttpCacheConfigParser implements ParserInterface
                     ->scalarNode('varnish_invalidate_token')
                         ->info('Optional: Varnish Invalidation token for purge')
                         ->defaultNull()
-                    ->end()
-                    ->scalarNode('varnish_bulk_purge')
-                        ->info('If purge should happen in bulk or not (if on varnish-module 0.9.x you should disable this)')
-                        ->defaultTrue()
                     ->end();
 
         foreach ($this->getExtraConfigParsers() as $extraConfigParser) {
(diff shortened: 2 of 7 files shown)


src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
```diff
@@ -37,6 +37,10 @@ public function addSemanticConfig(NodeBuilder $nodeBuilder)
                     ->scalarNode('varnish_invalidate_token')
                         ->info('Optional: Varnish Invalidation token for purge')
                         ->defaultNull()
+                    ->end()
+                    ->scalarNode('varnish_bulk_purge')
+                        ->info('If purge should happen in bulk or not (if on varnish-module 0.9.x you should disable this)')
+                        ->defaultTrue()
```
vidarl: If we add this setting, maybe it should default to `false`? and be set to `true` in 2.5.0 meta repo?

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -44,20 +45,20 @@ public function purge($tags)
...
-            $headers = [
-                'key' => $tag,
-                'Host' => empty($_SERVER['SERVER_NAME']) ? parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST) : $_SERVER['SERVER_NAME'],
-            ];
-
-            $headers = $this->addPurgeAuthHeader($headers);
+        $headers = $this->getPurgeHeaders();
+        $chunkSize = $this->determineTagsPerHeader($tags);
 
+        // NB! Thish requries varnish-moduls 0.10.2 or higher, 0.9.x only supported purging one tag at a time
```
vidarl: Have you confirmed that Platform.sh supports 0.10.2?

andrerom: No, but afaik varnish-moduels 0.9 won't compile on Varnish 5: https://github.com/varnish/varnish-modules/issues/44

src/DependencyInjection/ConfigResolver/HttpCacheConfigParser.php
andrerom: maybe, and that is one of the reason I'm leaning towards option B from PR description now.

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -96,4 +96,27 @@ private function addPurgeAuthHeader(array $headers)
...
+     *
+     * @return int
+     */
+    private function determineTagsPerHeader(array $tags)
+    {
+        if (!$this->configResolver->getParameter('http_cache.varnish_bulk_purge')) {
+            return 1;
+        }
+
+        if (mb_strlen(implode(' ', $tags)) < self::DEFAULT_HEADER_LENGTH) {
```
adamwojs: Nitpick: Could you please extract const representing tags separator? It's hardcoded in few places ;-)

andrerom:
@adamwojs & @vidarl updated

docs/varnish/vcl/varnish5.vcl
```diff
@@ -1,6 +1,6 @@
 // Varnish VCL for:
 // - Varnish 5.1 or higher (6.0LTS recommended, and is what we mainly test against)
-//   - Varnish xkey vmod (via varnish-modules package, or via Varnish Plus)
+//   - Varnish xkey vmod (via varnish-modules package 0,10.2 or higher, or via Varnish Plus)
```
kmadejski:
```suggestion
//   - Varnish xkey vmod (via varnish-modules package 0.10.2 or higher, or via Varnish Plus)
```

src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -44,20 +46,19 @@ public function purge($tags)
...
-            $headers = [
-                'key' => $tag,
-                'Host' => empty($_SERVER['SERVER_NAME']) ? parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST) : $_SERVER['SERVER_NAME'],
-            ];
-
-            $headers = $this->addPurgeAuthHeader($headers);
+        $headers = $this->getPurgeHeaders();
+        $chunkSize = $this->determineTagsPerHeader($tags);
 
+        // NB! This requries varnish-modules 0.10.2, if you need support for varnish-moduls 0.9.x, use ezplatform-http-cache 0.8.x
```
kmadejski:
```suggestion
        // NB! This requries varnish-modules 0.10.2, if you need support for varnish-modules 0.9.x, use ezplatform-http-cache 0.8.x
```

PR #86, opened 2019-05-15
[WIP] EZP-30562: As an Administrator I would like to use short HTTP cache tags

kmadejski:
| Question           | Answer
| **Type**           | Feature
| **Target version** | `0.9` / `master`
| **BC breaks**      | no*
| **Doc needed**     | yes

This PR introduces new `TagProvider` to get rid of hardcoded tags all around this bundle codebase. Moreover, shortening of the tags should help to solve recently reported Apache + PHP-FPM issues.

`ezplatform.http_cache.tag_provider` has been set as lazy, because in v1 there was an issue that `ConfigResolver` couldn't find the `http_cache.tag_format` parameter. 

`*` - _PR contains also `ShortToLongTagConverter` which has been made for BC, in other words, we don't want to force hard HTTP cache clear after the upgrade. It should be removed in the next major version, I think._

**TODO**:
- [x] Cover new Twig function `ez_httpcache_tag_*` in https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/using_tags.md
- [x] Ask for Code Review.

### DIFF
--- a/docs/using_tags.md
+++ b/docs/using_tags.md
@@ -20,30 +20,31 @@ Varnish or Fastly are highly recommended for medium to large traffic needs. Besi
 ## Tags in use in this bundle
 
 ### Tags for Content responses
+Tag format is configurable. By default, system uses `short` format, but you can use `long` format, for instance, for debugging purposes. 
 
-- `content-<content-id>` :
+- `c-<content-id>` / `content-<content-id>`:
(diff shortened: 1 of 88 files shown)


kmadejski:
Closed in favor of: https://github.com/ezsystems/ezplatform-http-cache/pull/112

PR #88, opened 2019-06-06
EZP-30627: Unified Twig helpers naming

pawbuj:
| Question           | Answer
| **JIRA issue**     | [EZP-30627](https://jira.ez.no/browse//EZP-30627)
| **Target version** |`master` 
| **BC breaks**      | yes
| **Doc needed**     | yes

related to https://github.com/ezsystems/ezplatform-admin-ui/pull/1008

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/Twig/ContentTaggingExtension.php
+++ b/src/Twig/ContentTaggingExtension.php
@@ -34,7 +34,7 @@ class ContentTaggingExtension extends Twig_Extension
     {
         return [
             new Twig_SimpleFunction(
-                'ez_http_tag_location',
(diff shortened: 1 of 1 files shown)


pawbuj:
This PR is part of biggest change , we want to clean the mess in the naming . We had a meeting and whole team decided that we make BC breaks. In our opinion it is not as problematic for developers to adapt the code to this change.

andrerom:
So there won't be possible to have code that works on both 2.5 and 3.0 then, not even templates?
cc @SylvainGuittard

PR #89, opened 2019-06-19
EZP-30544: Migrate Http Cache Bundle to FOS Cache Bundle v2

ViniTou:
| Question           | Answer
| **Type**           | Feature/Misc
| **Target version** |  `master`
| **BC breaks**      | yes
| **Doc needed**     | yes

## DESC

There are still some things to do, but before tackling them I have decided that this is the time (do'h finally) to show what have I done (; )

Mostly, those are updates in terms of BC and moving some stuff from us to FoS.

Reviewers, unleash all your horses on this.

**TODO**:
- [x] Fix tests + specs and passing (`$ composer test`)
- [x] Remove more stuff, that could be safely dropped (tbh, not sure what)
- [x] Ask for Code Review.

**Followups**
- Use ResponseTagger for REST
- (?) Use `PURGEKEYS` instead of `PURGE` in vcls


src/AppCache.php
```diff
@@ -6,34 +6,56 @@
...
+        return parent::fetch($request, $catch);
+    }
+
     /**
      * {@inheritdoc}
      */
     protected function createStore()
     {
-        return new TagAwareStore($this->cacheDir ?: $this->kernel->getCacheDir() . '/http_cache');
+        return new TagAwareStore([
```
andrerom: can be dropped, foshttpcache 2.x brings it's own

src/AppCache.php
```diff
@@ -6,34 +6,56 @@
...
+            'cache_tags_header' => 'xkey',
```
andrerom: we should switch to the default one by FOS

src/DependencyInjection/Compiler/HttpCachePass.php
```diff
@@ -22,42 +21,39 @@ class HttpCachePass implements CompilerPassInterface
...
     {
-        if (!$container->hasDefinition('ezplatform.http_cache.cache_manager')) {
+        if (!$container->hasDefinition('fos_http_cache.proxy_client.varnish.http_dispatcher')) {
             return;
         }
 
-        // Remove so 7.0 & 6.x kernels skip HttpCachePass
-        $container->removeAlias('ezpublish.http_cache.cache_manager');
+        $httpDispatcher = $container->findDefinition('fos_http_cache.proxy_client.varnish.http_dispatcher');
+        $httpDispatcher->replaceArgument(0, []);
```
andrerom: still needed? if so add comment on why it's here maybe, or make method name elf describing

src/EventSubscriber/XLocationIdResponseSubscriber.php
```diff
@@ -24,7 +24,7 @@ class XLocationIdResponseSubscriber implements EventSubscriberInterface
 {
     const LOCATION_ID_HEADER = 'X-Location-Id';
 
-    /** @var \FOS\HttpCache\Handler\TagHandler */
+    /** @var \EzSystems\PlatformHttpCacheBundle\Handler\TagHandler */
```
andrerom: Why do we need to type hint our own?

src/ProxyClient/Varnish.php
```diff
@@ -1,27 +1,41 @@
...
-use FOS\HttpCache\ProxyClient\Varnish as FOSVarnish;
+use FOS\HttpCache\ProxyClient\Varnish as FosVarnish;
 
-/**
- * Class Varnish.
- *
- * We need to support https as Platform.sh uses https by default.
- */
-class Varnish extends FOSVarnish
+class Varnish extends FosVarnish
```
andrerom: any reason why we can't drop this?

src/PurgeClient/LocalPurgeClient.php
```diff
@@ -22,7 +23,7 @@ class LocalPurgeClient implements PurgeClientInterface
      */
     protected $cacheStore;
 
-    public function __construct(RequestAwarePurger $cacheStore)
+    public function __construct(Psr6StoreInterface $cacheStore)
```
andrerom: Can we drop LocalPurgeClient?

src/Resources/config/fos_http_cache.yml
```diff
@@ -1,7 +1,10 @@
 proxy_client:
     default: varnish
     varnish:
-        servers: [$http_cache.purge_servers$]
+        http:
+            servers: ['$http_cache.purge_servers$']
+        tag_mode: 'purgekeys'
+        tags_header: 'key'
```
andrerom: todo: drop cusotm headers once VCL has been adapted

src/Resources/config/fos_http_cache.yml
```diff
@@ -13,4 +16,4 @@ user_context:
 tags:
     # Configure tag header for FosHttpCache
     # See docs/using_tags.md for further info on usage
-    header: '%ezplatform.http_cache.tags.header%'
+    response_header: '%ezplatform.http_cache.tags.header%'
```
andrerom: Todo: Drop custom header, same with `user_hash_header: X-User-Hash` once we have BC for old header being sent in responses.

src/ResponseTagger/Value/AbstractValueTagger.php
```diff
@@ -2,12 +2,12 @@
...
 namespace EzSystems\PlatformHttpCacheBundle\ResponseTagger\Value;
 
+use EzSystems\PlatformHttpCacheBundle\Handler\TagHandler;
 use EzSystems\PlatformHttpCacheBundle\ResponseTagger\ResponseTagger;
-use FOS\HttpCache\Handler\TagHandler;
 
 abstract class AbstractValueTagger implements ResponseTagger
 {
-    /** @var TagHandler */
+    /** @var \EzSystems\PlatformHttpCacheBundle\Handler\TagHandler */
```
andrerom: same here, any reason why we need to type hint our own?

andrerom:
@ViniTou Looks like very good progress, so get why you wanted input. Added some comments on things I think can be next steps to move further and remove some more abstraction. 

In terms of BC, the only one I think is relevant for most customers is `Var: X-User-Hash`, and this one we can handle in a response event subscriber to rewrite it to the header FOSHttpCache is using. kind of like we do for `X-Location-Id`.

src/AppCache.php
```diff
@@ -79,12 +80,7 @@ protected function invalidate(Request $request, $catch = false)
         }
 
         $response = new Response();
-        $store = $this->getStore();
-        if ($store instanceof RequestAwarePurger) {
-            $result = $store->purgeByRequest($request);
-        } else {
-            $result = $store->purge($request->getUri());
-        }
+        $result = $this->getStore()->purge($request->getUri());
```
andrerom: Where is tag invalidation done now? Or is it another verb used maybe.
I would expect this to not be able to handle tags as it's just the url.

Should we only pass to parent? Or would that mean we don't check IP (worth checking)

src/ProxyClient/Varnish.php
ViniTou: We could if we would like to move from `PURGE` to `PURGEKEYS` in Varnish.

src/PurgeClient/LocalPurgeClient.php
ViniTou: Both `LocalPurgeClient` and `VarnishPurgeClient` are now kept in sake of common interface for prefix decorator.

src/AppCache.php
ViniTou: I have _moved_ this into FOS provided Listeners.

src/ProxyClient/Varnish.php
andrerom: We probably should so we can align with FOS VCL as well as getting rid of the overloading, but this can be done in stages if you want.

src/PurgeClient/LocalPurgeClient.php
andrerom: ok, so keep for now until we can find another way to prefix tags

src/AppCache.php
ViniTou: used `\Toflar\Psr6HttpCacheStore\Psr6Store` directly

src/AppCache.php
ViniTou: switched to default settings

src/DependencyInjection/Compiler/HttpCachePass.php
ViniTou: I have changed code to make it more explicit and also added explanation for it.

src/EventSubscriber/XLocationIdResponseSubscriber.php
ViniTou: Fixed all over the place.

ViniTou:
@andrerom 
@alongosz 
@adamwojs 
@kmadejski 

If you can have another look at this, I would like to go with current scope of this PR and move to fastly, and then improve extra stuff related to varnish vcls.

.travis.yml
```diff
@@ -12,19 +12,21 @@ env:
...
-        - php: 7.1
-          env: TEST_CMD="bin/behat --profile=rest --tags=~@broken --suite=fullJson"
-        - php: 7.2
-          env: TEST_CMD="bin/phpunit -c vendor/ezsystems/ezplatform-rest/phpunit-integration-rest.xml" SYMFONY_CMD="ez:behat:create-language 'pol-PL' 'Polish (polski)'" PHP_IMAGE=ezsystems/php:7.2-v1
+        - name: "PHP 7.3 PHPUnit & PHPSpec"
+          php: 7.3
+        - name: "PHP 7.3 Rest Functional Tests"
+          php: 7.3
+          env:
+            - TEST_CMD="bin/phpunit --bootstrap vendor/autoload.php -c vendor/ezsystems/ezplatform-rest/phpunit-integration-rest.xml" SYMFONY_CMD="ez:behat:create-language 'pol-PL' 'Polish (polski)'"
```
andrerom: nit: while at it move `SYMFONY_CMD ` to own line as well

src/DependencyInjection/Compiler/KernelPass.php
```diff
@@ -46,7 +46,7 @@ protected function symfonyPre34BC(ContainerBuilder $container)
         $arguments = $container->getDefinition('cache_clearer')->getArguments();
 
         // BC Symfony < 3.4, as of 3.4 and up handles this itself, on lower versions we need to adjust the arguments manually
-        if (!is_array($arguments[0])) {
+        if (!\is_array($arguments[0])) {
```
adamwojs: I guess the whole `symfonyPre34BC` method might be removed 😉

src/PurgeClient/PurgeClientInterface.php
```diff
@@ -16,9 +16,9 @@ interface PurgeClientInterface
      *
      * It's up to the implementor to decide whether to purge tags right away or to delegate to a separate process.
      *
-     * @param array|int $tags Array of tags to purge, int for BC (location-<int>).
+     * @param array $tags Array of tags to purge.
      */
-    public function purge($tags);
+    public function purge(array $tags);
```
adamwojs:
```suggestion
    public function purge(array $tags): void;
```

src/Handler/TagHandler.php
```diff
@@ -20,39 +18,34 @@
...
-    public function purge($tags)
-    {
-        $this->purgeClient->purge($tags);
-    }
-
-    public function tagResponse(Response $response, $replace = false)
+    /**
+     * Tag a symfony response with the previously added tags.
+     *
+     * @param Response $response
```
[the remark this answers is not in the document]
ViniTou: As this phpdoc is from another package I have removed it.

Plopix:
just wondering @lserwatka @andrerom does that PR fix: https://jira.ez.no/browse/EZP-30518 ?
Thanks!

lserwatka:
@Plopix it might be fixed, new code does not use Host from globals, but always from configuration. We want to run some tests first to give more precise answer.

PR #92, opened 2019-07-08
EZP-30544: Refactored to use FOSHttpCache default ResponseHeader

ViniTou:
| Question           | Answer
| **Target version** | `master` for new features
| **BC breaks**      | yes
| **Doc needed**     | yes

Instead of relaying on `ezplatform.http_cache.tags.header` default fosCache tags are now used for local and varnish cache. 

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/DependencyInjection/EzPlatformHttpCacheExtension.php
+++ b/src/DependencyInjection/EzPlatformHttpCacheExtension.php
@@ -6,6 +6,7 @@
 namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection;
 
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ParserInterface;
+use FOS\HttpCache\TagHeaderFormatter\TagHeaderFormatter;
 use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Config\FileLocator;
@@ -35,6 +36,9 @@ class EzPlatformHttpCacheExtension extends Extension implements PrependExtension
         $loader->load('services.yml');
         $loader->load('event.yml');
         $loader->load('view_cache.yml');
+
+        $this->setDefaultResponseHeader($container);
+        $this->setDefaultTagSeparator($container);
     }
 
     public function prepend(ContainerBuilder $container)
@@ -63,5 +67,29 @@ class EzPlatformHttpCacheExtension extends Extension implements PrependExtension
     public function getExtraConfigParsers()
     {
         return $this->extraConfigParsers;
+    }
+
+    /**
+     * Overrides default header name based on setting tag_mode in FosHttpCacheBundle configuration.
+     */
+    private function setDefaultResponseHeader(ContainerBuilder $container): void
+    {
+        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+
+        $responseHeader = 'http' === $purgeType ? 'xkey' : TagHeaderFormatter::DEFAULT_HEADER_NAME;
+
+        $container->setParameter('fos_http_cache.tag_handler.response_header', $responseHeader);
+    }
+
+    /**
+     * Overrides default header separator based on setting tag_mode in FosHttpCacheBundle configuration.
+     */
+    private function setDefaultTagSeparator(ContainerBuilder $container): void
+    {
+        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+
+        $separator = 'http' === $purgeType ? ' ' : ',';
+
+        $container->setParameter('fos_http_cache.tag_handler.separator', $separator);
     }
 }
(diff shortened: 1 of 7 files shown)


andrerom:
Look ok, but be aware of intention/wish to simplify it again so tag formatting is more locally tied to proxy: https://github.com/FriendsOfSymfony/FOSHttpCache/issues/456#issuecomment-501631300

src/DependencyInjection/EzPlatformHttpCacheExtension.php
```diff
@@ -64,4 +68,28 @@ public function getExtraConfigParsers()
...
     }
+
+    /**
+     * Overrides default header name based on setting tag_mode in FosHttpCacheBundle configuration.
+     */
+    private function setDefaultResponseHeader(ContainerBuilder $container)
+    {
+        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+
+        $responseHeader = 'http' === $purgeType ? 'xkey' : TagHeaderFormatter::DEFAULT_HEADER_NAME;
```
kmadejski: Unsure about this, shouldn't it be something like the following? 
```
$responseHeader = in_array($purgeType, ['http', 'varnish', 'fastly']) ? 'xkey' : TagHeaderFormatter::DEFAULT_HEADER_NAME;
```
To cover all `http`-like possibilities? (ref: https://github.com/ezsystems/ezplatform/blob/master/config/packages/ezplatform.yaml#L25) Or did you drop these options and I'm not aware of it perhaps? 

The same question applies to https://github.com/ezsystems/ezplatform-http-cache/pull/92/files#diff-321f92f0967bd338c1a1e6404a0bb29fR91.

ViniTou: Yeah, I was not sure about those, `fastly` should probably taken care of in it own package.

Not sure about having dual configuration for `varnish/http`, from looking around the code and docs aside from comment above parameter we are using `http` as an valid purge type.  Should we definitely drop one or support both?

kmadejski: As for me, it's a good idea to take this opportunity and clean up these options. I'm good with keeping only `http`, but maybe there is some reason for keeping both. 
@andrerom do you have any input, perhaps?

andrerom: is `http` a generic enough option?
maybe we should rather end up with locale, varnish and fastly?
@vidarl Didn't you introduce the Varnish alias when you added Fastly in order to more clearly separate them?

ViniTou: Yes, I also think we should go with `local` (or maybe even rename it to `symfony` to be consistent with fos?), `fastly` and `varnish` as `xkey` is very specific to it, right?

>is http a generic enough option? Will it work on any kind of proxy? :)

I think yes, as long as it accepts PURGE requests and tag invalidation and expects `xkey`. But don't quote me on that ;)

src/DependencyInjection/EzPlatformHttpCacheExtension.php
```diff
@@ -64,4 +68,28 @@ public function getExtraConfigParsers()
     {
         return $this->extraConfigParsers;
     }
+
+    /**
+     * Overrides default header name based on setting tag_mode in FosHttpCacheBundle configuration.
+     */
+    private function setDefaultResponseHeader(ContainerBuilder $container)
```
adamwojs:
```suggestion
    private function setDefaultResponseHeader(ContainerBuilder $container): void
```

src/DependencyInjection/EzPlatformHttpCacheExtension.php
```diff
@@ -64,4 +68,28 @@ public function getExtraConfigParsers()
...
+
+        $responseHeader = 'http' === $purgeType ? 'xkey' : TagHeaderFormatter::DEFAULT_HEADER_NAME;
+
+        $container->setParameter('fos_http_cache.tag_handler.response_header', $responseHeader);
+    }
+
+    /**
+     * Overrides default header separator based on setting tag_mode in FosHttpCacheBundle configuration.
+     */
+    private function setDefaultTagSeparator(ContainerBuilder $container)
```
adamwojs:
```suggestion
    private function setDefaultTagSeparator(ContainerBuilder $container): void
```

src/DependencyInjection/EzPlatformHttpCacheExtension.php
andrerom: > (or maybe even rename it to symfony to be consistent with fos?),

if they call it that, I'm all for it.

> I think yes, as long as it accepts PURGE requests and tag invalidation and expects xkey. But don't quote me on that ;)

Afaik there is currently none, but I might be wrong ;)
LiteSpeed: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/444/files#diff-83872f222995ee0fbb9a54355f54c5c4R70
Fastly: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/451/files#diff-3584dcd2a732e6f34941b9194c2ff416R50

ViniTou: Then I am also for `symfony`, `varnish` and `fastly` but I would prefer to change those in separate PRs.

PR #93, opened 2019-07-09
EZP-30544: Use FOS default TagHeader and Purge Method for Varnish

ViniTou:
| Question           | Answer
| **BC breaks**      | yes
| **Doc needed**     | yes

- Drop custom `tags_header: key`, use default `xkey-softpurge`
- Use `PURGEKEYS` method instead of `PURGE`

**TODO**:
- [x] Ask for Code Review.


docs/varnish/vcl/varnish4.vcl
```diff
@@ -143,14 +143,14 @@ sub ez_purge {
         }
     }
 
-    if (req.method == "PURGE") {
+    if (req.method == "PURGE" || req.method == "PURGEKEYS") {
```
andrerom: varnish4.vcl can be dropped while on it, Varnish 4 support was dropped already in 2.5, but the file is here as 0.9 is also meant to work with 1.13. For master you can drop it.

src/DependencyInjection/Compiler/HttpCachePass.php
```diff
@@ -22,39 +26,34 @@ class HttpCachePass implements CompilerPassInterface
...
-        $container->removeDefinition('fos_http_cache.proxy_client.varnish.http_dispatcher');
-    }
-
-    private function processVarnishProxyClient(ContainerBuilder $container)
     {
         if (!$container->hasDefinition('fos_http_cache.proxy_client.varnish')) {
             throw new InvalidArgumentException('Varnish proxy client must be enabled in FOSHttpCacheBundle');
         }
+        // Override FOS default httpDispatcher with \EzSystems\PlatformHttpCacheBundle\ProxyClient\HttpDispatcherFactory.
+        $container->removeDefinition('fos_http_cache.proxy_client.varnish.http_dispatcher');
```
alongosz: It's rather recommended to decorate.

src/DependencyInjection/Compiler/HttpCachePass.php
```diff
@@ -22,39 +26,34 @@ class HttpCachePass implements CompilerPassInterface
...
 
         $fosConfig = array_merge(...$container->getExtensionConfig('fos_http_cache'));
 
         $servers = $fosConfig['proxy_client']['varnish']['http']['servers'] ?? [];
         $baseUrl = $fosConfig['proxy_client']['varnish']['http']['base_url'] ?? '';
```
alongosz: Do we have these exposed as parameters? If not, then ok, decorating might not be possible in a clean way.

PR #94, opened 2019-07-11
EZP-30544: Use default user hash header name

ViniTou:
| Question           | Answer
| **BC breaks**      | yes
| **Doc needed**     | yes

Use default `X-User-Context-Hash` header name

related:
https://github.com/ezsystems/ezpublish-kernel/pull/2696
https://github.com/ezsystems/ezplatform-page-fieldtype/pull/99

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/docs/fos_http_cache.md
+++ b/docs/fos_http_cache.md
@@ -13,7 +13,7 @@ As HttpCache in eZ Publish/Platform predates FOSHttpCache, we have own abstracti
  - App Cache: `EzSystems\PlatformHttpCacheBundle\AppCache` app cache class to extend from, this in turn extends
    `FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache`, however it provides own set of request-aware purgers and
    own handling of user context hash which for BC reasons is called `X-User-Hash` in eZ.
- - From version 1.0, FosCache default is used `X-User-Context-Hash` for user context hash header name.
+ - Since version 1.0, FOSHttpCache's `X-User-Context-Hash` header is used for user context hash.
 
 _Note: Once FOSHttpCache(Bundle) has full support for tagging, a major version of this bundle might be be refactored to more
 directly reuse FOSHttpCache(Bundle). However in the meantime our abstractions have allowed us to ship features across
--- a/docs/fos_http_cache.md
+++ b/docs/fos_http_cache.md
@@ -13,6 +13,7 @@ As HttpCache in eZ Publish/Platform predates FOSHttpCache, we have own abstracti
  - App Cache: `EzSystems\PlatformHttpCacheBundle\AppCache` app cache class to extend from, this in turn extends
    `FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache`, however it provides own set of request-aware purgers and
    own handling of user context hash which for BC reasons is called `X-User-Hash` in eZ.
+ - From version 1.0, FosCache default is used `X-User-Context-Hash` for user context hash header name.
 
 _Note: Once FOSHttpCache(Bundle) has full support for tagging, a major version of this bundle might be be refactored to more
 directly reuse FOSHttpCache(Bundle). However in the meantime our abstractions have allowed us to ship features across
--- a/src/EventSubscriber/RequestEventSubscriber.php
+++ b/src/EventSubscriber/RequestEventSubscriber.php
@@ -12,7 +12,7 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 
-class RequestEventSubscriber implements EventSubscriberInterface
+final class RequestEventSubscriber implements EventSubscriberInterface
 {
     /** @var string */
     private $userHashHeaderName;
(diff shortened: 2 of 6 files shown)


andrerom:
What is the ´RequestEventSubscriber´ for? I was partly expecting an event subscriber to rewrite headers using old header name, but this seems to do something else right?

ViniTou:
@andrerom 
it should be more clear if you look at https://github.com/ezsystems/ezpublish-kernel/pull/2696. Basically it setups headers that should be forwarded.

src/EventSubscriber/RequestEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+        if ($event->isMasterRequest()) {
+            $request = $event->getRequest();
+
+            if (
+                $request->attributes->get('needsForward') &&
+                $request->attributes->has('semanticPathinfo') &&
+                $request->headers->has($this->userHashHeaderName)
+            ) {
+                $headersToForward = $request->attributes->get('forwardRequestHeaders', []);
+                $request->attributes->set('forwardRequestHeaders', array_merge(
```
adamwojs: I would replace `array_merge` call to:

```php
$headersToForward = $request->attributes->get('forwardRequestHeaders', []);
$headersToForward[$this->userHashHeaderName] = $request->headers->get($this->userHashHeaderName)

$request->attributes->set('forwardRequestHeaders', $headersToForward[);
```

src/EventSubscriber/RequestEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+declare(strict_types=1);
+
+namespace EzSystems\PlatformHttpCacheBundle\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class RequestEventSubscriber implements EventSubscriberInterface
```
adamwojs:
```suggestion
final class RequestEventSubscriber implements EventSubscriberInterface
```

src/EventSubscriber/RequestEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /** @var string */
+    private $userHashHeaderName;
+
+    public function __construct(string $userHashHeaderName)
+    {
+        $this->userHashHeaderName = $userHashHeaderName;
+    }
+
+    public static function getSubscribedEvents()
```
adamwojs:
```suggestion
    public static function getSubscribedEvents(): array
```

src/EventSubscriber/RequestEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        return [
+            KernelEvents::REQUEST => [
+                ['onKernelRequestForward', 15],
+            ],
+        ];
+    }
+
+    public function onKernelRequestForward(GetResponseEvent $event)
```
adamwojs:
```suggestion
    public function onKernelRequestForward(GetResponseEvent $event): void
```

andrerom:
> But maybe you had something else in mind for this to be cleaner?

If we'd like to avoid silently break controllers setting vary by `X-User-Hash` we could have swapped over the value to new header in a late response event. This is what I was referring to.

src/Resources/config/view_cache.yml
```diff
@@ -40,6 +40,12 @@ services:
         tags:
             - { name: kernel.event_subscriber }
 
+    EzSystems\PlatformHttpCacheBundle\EventSubscriber\RequestEventSubscriber:
+        arguments:
+            - "@=container.getParameter('fos_http_cache.event_listener.user_context.options')['user_hash_header']"
```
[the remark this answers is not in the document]
ViniTou: Does this relates to your worries?
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/FOSHttpCacheExtension.php#L298

alongosz:
> > But maybe you had something else in mind for this to be cleaner?
> 
> If we'd like to avoid silently break controllers setting vary by `X-User-Hash` we could have swapped over the value to new header in a late response event. This is what I was referring to.

But then we still would have this coupling of kernel to previous version of HTTP Cache (because we'd still need to have `X-User-Hash` in Kernel. I'd rather say that this needs to be documented as something that changed. 

Relying on `X-User-Hash` in my controller, only to see later on a response `X-User-Context-Hash` would actually be more confusing for me. 

Moreover if you want to avoid silent breaks, you probably need to have both `X-User-Hash` and `X-User--Context-Hash` on the response, which gets even more confusing.

The architecture here is as follows: ezplatform-http-cache depends on ezpublish-kernel. ezpublish-kernel exposes extension point to forward some specific request headers, ezplatform-http-cache uses that extension point to forward `X-User-Context-Hash`. I'm not very happy with an extension point in a Subscriber, but right now I don't see a good and decoupled alternative.

docs/fos_http_cache.md
```diff
@@ -13,6 +13,7 @@ As HttpCache in eZ Publish/Platform predates FOSHttpCache, we have own abstracti
  - App Cache: `EzSystems\PlatformHttpCacheBundle\AppCache` app cache class to extend from, this in turn extends
    `FOS\HttpCacheBundle\SymfonyCache\EventDispatchingHttpCache`, however it provides own set of request-aware purgers and
    own handling of user context hash which for BC reasons is called `X-User-Hash` in eZ.
+ - From version 1.0, FosCache default is used `X-User-Context-Hash` for user context hash header name.
```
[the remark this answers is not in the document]
DominikaK: The suggestion sounds good 👍

PR #95, opened 2019-07-15
EZP-30662: Added http cache tests

mnocon:
| Question           | Answer
| **JIRA issue**     |  https://jira.ez.no/browse/EZP-30662
| **Type**           | Misc (Tests)
| **Target version** |  master
| **BC breaks**      | no
| **Doc needed**     | no

This PR adds basic Behat configuration, together with first scenarios (for now using Symfony cache, as 3.0.x does not support running Varnish on Travis - see https://jira.ez.no/browse/EZP-30750)

Making sure that the headers are correct and that cache is refreshed when a Content Item is edited.

Related:
ezsystems/ezplatform#436
ezsystems/BehatBundle#89


features/symfony/cache.feature
```diff
@@ -0,0 +1,52 @@
...
+    And I am viewing the pages on siteaccess "site" as "<user>" "<password>"
+    And I visit "<itemName>" on siteaccess "site"
+    And I see correct preview data for "Folder" Content Type
+      | field | value      |
+      | title | <itemName> |
+    When I edit "<itemName>" Content item in "eng-GB"
+      | short_name          |
+      | <itemNameAfterEdit> |
+    And I reload the page
+    And I reload the page
```
m-tyrala: Why there's double refresh here?

mnocon: The cache is purged after Content Item is edited. 
First request to the server results in creating the cache (and headers: `X-Symfony-Cache: GET /foldertest: miss, store` in response), second request should have the response served from the cache. As we want to test the caching behaviour I've decided to check that the second request indeed has `GET /mysite: fresh` response header

mnocon:
@lserwatka this one can be merged

PR #96, opened 2019-08-04
EZP-30814: Removed Twig deprecations

adamwojs:
| Question           | Answer
| **JIRA issue**     | [EZP-30814](https://jira.ez.no/browse/EZP-30814)
| **BC breaks**      | no
| **Doc needed**     | no

Continuation of https://github.com/ezsystems/ezpublish-kernel/pull/2708:

* Replaced `spaceless` tag usages in favour of `spaceless` filter
* Removed `for ... if ` statements

**TODO**:
- [ ] ~Implement tests + specs and passing (`$ composer test`)~
- [X] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [ ] Ask for Code Review.

### DIFF
--- a/src/Resources/views/fields/content_fields.html.twig
+++ b/src/Resources/views/fields/content_fields.html.twig
@@ -6,11 +6,13 @@
     {% apply spaceless %} %}
         {% if not ez_field_is_empty( content, field ) %}
             <ul {{ block( 'field_attributes' ) }}>
-                {% for contentId in field.value.destinationContentIds if parameters.available[contentId] %}
-                     {{ fos_httpcache_tag('relation-' ~ contentId) }}
-                    <li>
-                        {{ render( controller( "ez_content:viewAction", {'contentId': contentId, 'viewType': 'embed', 'layout': false} ) ) }}
-                    </li>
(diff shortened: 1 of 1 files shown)


andrerom:
applicable to stable branches as well to get rid of deprecation notices?

adamwojs:
@andrerom "Yes" in general, but not applicable in this case: `src/Resources/views/fields/content_fields.html.twig` doesn't exist on 0.9.X.

src/Resources/views/fields/content_fields.html.twig
```diff
@@ -3,22 +3,24 @@
 {% extends "EzPublishCoreBundle::content_fields.html.twig" %}
 
 {% block ezobjectrelationlist_field %}
-    {% spaceless %}
+    {% apply spaceless %} %}
```
emodric: There's an extra `%}` here and in other places below.

adamwojs: Good catch, thank you! Follow up PR: https://github.com/ezsystems/ezplatform-http-cache/pull/97

PR #99, opened 2019-08-20
[WIP] Install latest dependencies

mnocon:
| Question           | Answer
| **JIRA issue**     | [EZP-XXXXX](https://jira.ez.no/browse/EZP-XXXXX)
| **Type**           | Bug/Improvement/Feature/Misc
| **Target version** | latest stable `0.8` for bug fixes, `master` for new features
| **BC breaks**      | yes/no
| **Doc needed**     | yes/no

**TODO**:
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [ ] Ask for Code Review.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,6 @@
         "friendsofphp/php-cs-fixer": "~2.15.0",
         "memio/spec-gen": "^0.9"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "autoload": {
         "psr-4": {
             "EzSystems\\PlatformHttpCacheBundle\\": "src",
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
         }
     ],
     "require": {
+        "ezsystems/doctrine-dbal-schema": "^1.0@dev",
         "ezsystems/ezpublish-kernel": "^8.0@dev",
         "ezsystems/ezplatform-rest": "^1.0@dev",
         "friendsofsymfony/http-cache-bundle": "^2.6.1",


composer.json
```diff
@@ -10,6 +10,7 @@
         }
     ],
     "require": {
+        "ezsystems/doctrine-dbal-schema": "^1.0@dev",
```
andrerom: Do we use `doctrine-dbal-schema` within this package?

mnocon: We don't, directly. This is advised by Composer doc:
```
If one of your dependencies has a dependency on an unstable package you need to explicitly require it as well, along with its sufficient stability flag.
```
`doctrine-dbal-schema:1.0` (a dependency of ezpublish-kernel) exists as `dev` and `beta` versions, meaning no stable versions.

Problem is, that by using:
```
    "minimum-stability": "dev",
    "prefer-stable": true,
```
Composer will install ezpublish-kernel:8.0.0-beta1, which is quite outdated now (for example. all the Events refactoring happened after that)

When using just `minimum-stability": "dev",`we would get all Symfony dev dependencies (too many of them).

Without specyfing `doctrine-dbal-schema` (and removing the prefer-stable, minimum stability options) executing `composer install` won't work for this package.

See: https://github.com/ezsystems/ezplatform-richtext/pull/57 , where the same issue happened (and also many other packages).

I've added prefer-stable, minimum stability options when work for 3.0 has started as it worked then, but when we've released the beta it stopped working as we'd like. 

My idea is to remove the unneeded dependencies before we release 3.0 final, I don't like this solution too much to be honest, so all suggestions are welcome.

alongosz:
Resolved via 071eaacdc4fab02fc9186e64b175d97a422802ae.

Issue #79, opened 2019-03-06
Tags can reach header limits on Varnish & Fastly


andrerom:
Varnish header limit: [8k](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-resp-hdr-len) _(configurable)_ _<-- There has been cases with customers reaching this in 2018_
Fastly header limit: [16k](https://docs.fastly.com/guides/debugging/resource-limits#request-and-header-limits) _(**not** configurable)_

Tags should ideally be shortened in a future version of this package to reduce risk of hitting the limits on complex pages _(Landing Pages, ...)_.

e.g. _(depends on how much we want to favour readability)_:
- `c-<content-id>` 
- `c-versions-<content-id>`
- `c-type-<content-type-id>`
- `l-<location-id>`
- `prnt-<parent-location-id|location-id>`
- `pth-<location-id>`
- `rl-<content-id>`
- `rl-l-<location-id>`

Compared to what we have today: https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/using_tags.md#tags-in-use-in-this-bundle

**Backwards Compatibility**: Bundle can still purge the old tags in case someone tags header with old tags in their code, but this will be a break for any custom purge logic _(there is no official way to do that at time of writing)_

PR #104, opened 2019-10-08
EZP-30562: Make it possible to ignore tags over configurable max length

ITernovtsiy:
| Question           | Answer
| **Type**           |  Feature
| **Target version** | `0.9` / `master`
| **BC breaks**      | no
| **Doc needed**     | yes

related to 
https://github.com/ezsystems/ezplatform-http-cache/pull/86 
https://github.com/ezsystems/ezplatform-http-cache/issues/79
While using shortened tags will help to "reduce the risk" of 500 error, it still will be possible to embed too many objects in admin and got a 500 error page.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/docs/using_tags.md
+++ b/docs/using_tags.md
@@ -9,7 +9,7 @@ They work in a similar way as [persistence cache tags in eZ Platform v2](https:/
 
 It works across all supported proxies _(see ["drivers"](drivers.md))_ by eZ Platform:
 - Symfony Proxy _(PHP based for single server usage, primarily for smaller web sites)_
-- [Varnish](https://varnish-cache.org/) with [xkey module](https://github.com/varnish/varnish-modules) _or_ [Varnish Plus](https://www.varnish-software.com/products/varnish-plus/) _(High performance reverse proxy)_
+- [Varnish](https://varnish-cache.org/) with [xkey module (0.10.2+)](https://github.com/varnish/varnish-modules) _or_ [Varnish Plus](https://www.varnish-software.com/products/varnish-plus/) _(High performance reverse proxy)_
 - [Fastly](https://www.fastly.com/) _(High performance reverse proxy, originally based on Varnish, worldwide as a CDN, driver available in eZ Platform Enterprise)_
 
 _In order to support several repositories on one installation, tags will be prefixed by
--- a/docs/using_tags.md
+++ b/docs/using_tags.md
@@ -172,4 +172,13 @@ One common issue to encounter is that the tagging headers exceed limits in Varni
 - [http_max_hdr](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-max-hdr) (e.g. 128)
 - [http_resp_size](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-resp-size) (e.g. 64k)
 
+In case you exceed the limit on Nginx, 500 error will be returned.
+Consider using fewer tags or ignore tags over the limit. It can be done by setting a max length parameter (in bytes)
+```yaml
+parameters:
+    # Warning, setting this means you risk losing tag information, which means you will end up with stale cache
+    # First tune your system if you can, i.e. nginx: proxy_buffer_size, Varnish: http_resp_hdr_len, on Fastly limit is 16k
+    ezplatform.http_cache.tags.header_max_length: 16000
+```
+
 For more up-to-date info see online doc: https://doc.ezplatform.com/en/latest/guide/http_cache/#available-tags
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -28,6 +28,8 @@ class TagHandler extends FOSTagHandler
     private $tagsHeader;
     /** @var int|null */
     private $tagsHeaderMaxLength;
+    /** @var int|null */
+    private $tagsHeaderReducedTTl;
     /** @var LoggerInterface */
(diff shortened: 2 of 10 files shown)


ITernovtsiy:
@andrerom please review when you'll have a chance, thanks.

andrerom:
Status: This is part of a bigger problem, we _also_ need to find a good way to reduce size of tags, and if they go above a certain size _(i.e. 8kb, which will cause issues several places)_, log warning and shorten.

I started some time ago to try to cleanup #86, however I now think there are two options to consider:
- Aim for new 0.10 release where all tags are shortened, to more or less what #86 proposes, no complexity trying to support both at the same time
- Like Drupal(?): Keep tags as is and use a hashing algo in production that produces shorter strings, strings doesn't need to be 100% unique, but as close to it as possible with as short strings as possible would be good. Can even have setting for this

--

So in summary: We can move forward on this, but as soon as we have smaller tags we should maybe do a follow up on this to set a sane default.

src/Handler/TagHandler.php
```diff
@@ -57,19 +59,20 @@ public function tagResponse(Response $response, $replace = false)
         $tags = [];
         if (!$replace && $response->headers->has($this->tagsHeader)) {
             $headers = $response->headers->get($this->tagsHeader, null, false);
-            if (!empty($headers)) {
+            if ($headers) {
```
andrerom: irrelevant change?

src/Handler/TagHandler.php
```diff
@@ -57,19 +59,20 @@ public function tagResponse(Response $response, $replace = false)
...
                 // handle both both comma (FOS) and space (this bundle/xkey/fastly) separated strings
                 // As there can be more requests going on, we don't add these to tag handler (ez-user-context-hash)
                 $tags = preg_split("/[\s,]+/", implode(' ', $headers));
             }
         }
 
         if ($this->hasTags()) {
+            // note that 'getTagsHeaderValue' returns unique tags
```
andrerom: yes, but after merge they are not guaranteed to be unique right?

src/Handler/TagHandler.php
ITernovtsiy: yes, just a micro-optimization. I think it worsts it here, as this function called on every request. 
Let me know if I should remove it.

src/Handler/TagHandler.php
ITernovtsiy: right, missed that, fixed

src/Handler/TagHandler.php
andrerom: Then I think you should remove it.

Reason besides being irrelevant to the change here: afaik `empty` is faster, or _at least_ as fast as changing to evaluating if the value is true or false. `empty` like `isset` is language constructs, and not functions, so they don't have function overhead.

docs/using_tags.md
```diff
@@ -172,4 +172,11 @@ One common issue to encounter is that the tagging headers exceed limits in Varni
 - [http_max_hdr](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-max-hdr) (e.g. 128)
 - [http_resp_size](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-resp-size) (e.g. 64k)
 
+In case you exceed the limit on Nginx, 500 error will be returned.
+Consider using fewer tags or ignore tags over the limit. It can be done by setting a max length parameter (in bytes)
+```yaml
+parameters:
+    ezplatform.http_cache.tags.header_max_length: 4000
```
andrerom:
```suggestion
    # Warning, setting this means you risk losing tag information, which means you can end up with stale cache
    # First tune your system if can, i.e. nginx: proxy_buffer_size, Varnish: http_resp_hdr_len, on Fastly limit is 16k
    ezplatform.http_cache.tags.header_max_length: 16000
```

src/Handler/TagHandler.php
ITernovtsiy: Ok, reverted original code here and in `if (!empty($repoPrefix)) {`.

regarding `empty` vs `bool cast`, bool cast is faster (both with and without opcache optimization).
Proof:
empty_vs_boolcast
`!empty` variant performs 2 extra operations without opcache (just 1 extra after opcache optimization).

andrerom: I stand corrected :)

src/Handler/TagHandler.php
```diff
@@ -80,7 +86,14 @@ static function ($tag) use ($repoPrefix) {
                 $tags[] = 'ez-all';
             }
 
-            $response->headers->set($this->tagsHeader, implode(' ', array_unique($tags)));
+            $tagsString = implode(' ', $tags);
+            if ($this->tagsHeaderMaxLength && strlen($tagsString) > $this->tagsHeaderMaxLength) {
+                $tagsString = trim(substr($tagsString, 0, strrpos(
```
andrerom: I'd like us to add a warning here when we end up dropping tags.

ITernovtsiy: Done, warning message:
HTTP Cache tags header max length reached and truncated to `$this->tagsHeaderMaxLength`
Let me know if it should be updated.

ITernovtsiy:
@andrerom I can close this PR, right? I see this functionality released with v2.5.9

andrerom:
Yes indeed, thanks :)

PR #106, opened 2019-11-07
EZP-31113: AssignSectionToSubtreeSignal is not handled by search and HTTP cache

kmadejski:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-31113](https://jira.ez.no/browse/EZP-31113)
| **Type**           | Bug
| **Target version** | `0.9`
| **BC breaks**      | no
| **Doc needed**     | no

This PR depends on: https://github.com/ezsystems/ezpublish-kernel/pull/2862. Please refer to it for more information.

**TODO**:
- [x] Implement feature / fix a bug.
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.
- [ ] Drop changes on merge to `master`
- [x] Create a separated PR which uses events instead (for `master`).

PR #107, opened 2019-11-27
EZP-31113: AssignSectionToSubtree event is not handled by search and HTTP cache

adamwojs:
| Question           | Answer
| **JIRA issue**     | [EZP-31113](https://jira.ez.no/browse/EZP-31113)
| **Type**           | Bug
| **BC breaks**      | no
| **Doc needed**     | no

https://github.com/ezsystems/ezplatform-http-cache/pull/106 for master

**TODO**:
- [x] Rebase after merge https://github.com/ezsystems/ezplatform-http-cache/pull/108
- [x] Ask for Code Review.

### DIFF
--- a/src/EventSubscriber/CachePurge/LocationEventsSubscriber.php
+++ b/src/EventSubscriber/CachePurge/LocationEventsSubscriber.php
@@ -16,6 +16,7 @@ use eZ\Publish\API\Repository\Events\Location\MoveSubtreeEvent;
 use eZ\Publish\API\Repository\Events\Location\SwapLocationEvent;
 use eZ\Publish\API\Repository\Events\Location\UnhideLocationEvent;
 use eZ\Publish\API\Repository\Events\Location\UpdateLocationEvent;
+use eZ\Publish\API\Repository\Events\Section\AssignSectionToSubtreeEvent;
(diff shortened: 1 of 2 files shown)


adamwojs:
> Just a side note: I believe it's enough to invalidate path-* only but there is a case I'm not aware of, perhaps 🙂

I will check it 😉

andrerom:
> I will check it 😉

I can confirm that unless some views are missing path tagging.

As for parent, any reason why we think parent should be cleared too?
Could make sense, but then it would be to make sure parent view (not siblings) are cleared in case it inlines children listings for instance, but don't add tags for children for some reason (so bug?).

micszo:
Status update. The Solr part works fine on v3. However there are doubts if Varnish (6.1) works correctly with v3 (master, w/o these changes). @webhdx is helping me confirm if it's a configuration issue or if perhaps we have a bug somewhere. I'll let you know when we pinpoint something.

micszo:
Good news! It appears that https://github.com/ezsystems/ezplatform-http-cache/pull/118 resolves the issue I mentioned before. 🎉 So this finally might be unblocked. \cc @webhdx

PR #110, opened 2020-01-02
[WIP] EZP-30562: Hash tags with CRC32 in PROD to reduce cache tag sizes

andrerom:
| Question           | Answer
| **Target version** | 0.9
| **BC breaks**      | no
| **Doc needed**     | yes

This suggests to replace #86 for the following reasons:
- That one introduces shorter tags, but keeps old tags as well for BC which complicates the code a lot
    - Even with shorter tags they still continue to be rather long with repository prefix
- This is inspired by cache plugin for Wordpress, simply hashing the tag to keep it short
   - Picked CRC32 for this as it's 1. short 2. pretty fast 3. seems to be the one with short hash that avoids collisions the most _(some referred to it as 1 in 4 billion, at such level we don't care if we might sometimes in a full moon end up clearing a little bit too much cache as tag length is more of an issue..)_ 
   - Introduces setting for this and enables by default in prod, so dev continues to be readable _(as dev is not cached I assume this should be safe)_

WDYT @kmadejski @ITernovtsiy ?

Related to #104
Closes #86

**TODO**:
- [ ] 🗣 OPEN TOPIC: Should we adapt to also *purge* non hashed tags in 0.9 for BC? Otherwise we need to make it more clear people must clear **all** cache when deploying this change _(enabled)_
- [ ] Add Doc here and on doc.ezplatform.no
- [ ] Add tests + specs and passing (`$ composer test`)
- [ ] Functional testing
- [x] Ask for Code Review.

### DIFF
--- a/spec/Handler/TagHandlerSpec.php
+++ b/spec/Handler/TagHandlerSpec.php
@@ -29,16 +29,16 @@ class TagHandlerSpec extends ObjectBehavior
         $this->beConstructedWith($cacheManager, 'xkey', $purgeClient, $tagPrefix);
     }
 
-    public function it_calls_purge_on_invalidate()
+    public function it_calls_purge_client_on_invalidate(PurgeClientInterface $purgeClient)
     {
-        $this->purge(Argument::exact(['something']));
+        $purgeClient->purge(Argument::exact(['something']))->shouldBeCalled();
 
         $this->invalidateTags(['something']);
     }
 
     public function it_calls_purge_client_on_purge(PurgeClientInterface $purgeClient)
     {
-        $purgeClient->purge(Argument::exact(['something']));
+        $purgeClient->purge(Argument::exact(['something']))->shouldBeCalled();
 
         $this->purge(['something']);
     }
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -26,29 +26,36 @@ class TagHandler extends FOSTagHandler
     private $purgeClient;
     private $prefixService;
     private $tagsHeader;
+    private $hashTags;
 
     public function __construct(
         CacheManager $cacheManager,
         $tagsHeader,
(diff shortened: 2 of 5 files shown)


ITernovtsiy:
@andrerom I like this solution.
The only concern is that it will produce longer tags in some cases (comparing to #86 ).
But I'm not sure if it worth to have both solutions in place and hash only if short tag length > 8. Probably, it's better to keep this simple and use only hash.

andrerom:
> The only concern is that it will produce longer tags in some cases (comparing to #86 ).

Yes, compared to `c-<id>` it will for the first 999 content object ids produce a bit longer hash.
But it will *always*\* be smaller then what we have now, and on `pa-<id>`  _(the one there will be most of on each and every content response)_ it will only be longer for the first 99 locations.


So as you say we can combine them and make it detect length before hashing, but downside would be:
- break / complexity of introducing new tags
- readability in dev, the shorter once are making it far less obvious what they represent

\* Exception is `path-<id>`, it will be longer when id is 1-9... :P

andrerom:
But I'm open for that.

Maybe the best way to decide is to ask if people that run into this issue can evaluate if this would solve it for them.
In your case, whats your limit? and where?

ITernovtsiy:
I think my case is not applicable here. We have table with ~2k embeds, it will go over our 4k limit even with the shortest possible tag).
So, I would vote for using just `hash`, disregard length.

andrerom:
For that case maybe find a way to drop all path tags coming from embeds, and reduce TTL time a bit for the remote chance sub tree operations affect embeds but not the content embedding them.

andrerom:
Maybe we can do something like that by default when detecting to much tags, drop what we think is less prone to being affected and revise ttl.

andrerom:
Closing as after some discussions figured out we can have even shorter tags, and it will be more readable than this.

PR #112, opened 2020-01-28
[1.0] Reduce Cache tag size by 40-70% & expand doc on how to deal with tags exceeding limits

andrerom:
| Question           | Answer
| **Type**           | Improvement/Feature/Bug
| **Target version** | new `1.0` branch for LTS releases, implies master should become `2.0`
| **BC breaks**      | yes
| **Doc needed**     | yes

#### Why

> Servers like Apache or nginx have a limit of a response header size. In some cases, for instance, when RichText contains a lot of embedded images, HTTP tags header (x-key) might be too long. It might cause 500 error in case of environment based on Apache + PHP-FPM. To ~avoid~ reduce such a situation tags should be shortened.

#### What
_Builds on what @kmadejski proposed in #86 after discussions in PS recently._

Reduces tags from i.e. `intranet_content-22` to `1c22`, compared to recent proposal to hash tags this:
- Is mostly shorter _(with hash it would always be 8 characters)_
- More readable, and does not differ between `dev` & `prod`

BC is kept on purges in content slots for this branch, so templates/php code using documented tags already won't stop working, but this is done as own commit so it can be reverted for v3 branch.

**TODO**:
- [x] Ask for Code Review.
- **Doc**: Update doc.ezplatform.com once this is accepted and switched to in meta branch
- **DO NOT MERGE THIS**: Once approved and cleaned up just close PR and let branch stays as is, merge to master and bump master to 2.x.

### DIFF
--- a/docs/using_tags.md
+++ b/docs/using_tags.md
@@ -13,60 +13,71 @@ It works across all supported proxies _(see ["drivers"](drivers.md))_ by eZ Plat
 - [Fastly](https://www.fastly.com/) _(High performance reverse proxy, originally based on Varnish, worldwide as a CDN, driver available in eZ Platform Enterprise)_
 
 _In order to support several repositories on one installation, tags will be prefixed by
-repository name on non default repositories. E.g. "intranet_path-1"._
+repository index in use. I.e. 0p1", where "0" is the repository prefix._
 
-Varnish or Fastly are highly recommended for medium to large traffic needs. Besides being able to handle much more traffic, supported for use in cluster setup, they also both support soft purge _(by tags)_, meaning they are able to serve stale content while it's refreshed in the background on-demand, leading to more stable load on your backend.
+Varnish or Fastly are highly recommended for medium to large traffic needs. Besides being able to handle much more traffic,
+supported for use in cluster setup, they also both support soft purge _(by tags)_, meaning they are able to serve stale
+content while it's refreshed in the background on-demand, leading to more stable load on your backend.
+
 
 ## Tags in use in this bundle
 
 ### Tags for Content responses
 
-- `content-<content-id>` :
+- *Content*:\
+  `c<content-id>`:
   _Used on anything that is affected by changes to content, on content itself as well as location and so on._
 
-- `content-versions-<content-id>` :
-  _Used for clearing cache for content version list views, when not affecting the published content._
+- *Content Version*:\
+  `cv<content-id>`:
+  _Used for clearing cache for content version list views, when *not* affecting the published content._
 
-- `content-type-<content-type-id>` :
+- *Content Type*:\
+  `ct<content-type-id>`:
   _For use when content type changes affecting content of its type._
 
-- `location-<location-id>` :
+- *Location*:\
+  `l<location-id>`:
   _Used for clearing all cache relevant for a given location._
 
-- `parent-<parent-location-id>` :
-  _Used for clearing cache of all siblings of an location._
-
-- `parent-<location-id>` :
+- *Parent Location*:\
+  For responses this represent the parent location, on purges it's used in 2 distinct ways:\
+  `pl<parent-location-id>`:
+  _Used for clearing cache of all siblings of an location._\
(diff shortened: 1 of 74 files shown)


docs/using_tags.md
```diff
@@ -149,27 +160,64 @@ E.g. on Move Location signal the following tags will be purged:
...
+
+*Varnish* config:
+- [http_resp_hdr_len](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-resp-hdr-len) (default 8k, change to i.e. 32k)
+- [http_max_hdr](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-max-hdr) (default 64, change to i.e. 128)
+- [http_resp_size](https://varnish-cache.org/docs/6.0/reference/varnishd.html#http-resp-size) (default 23k, change to i.e. 96k)
+- [workspace_backend](https://varnish-cache.org/docs/6.0/reference/varnishd.html#workspace-backend) (default 64k, change to i.e. 128k)
+
+*Fastly* has a `Surrogate-Key` header limit of *16kb*, this can *not* be changed.
+
+*Apache* has a limit of 8kb, and seems to need to be re-compiled to change this TODO: Find link about this
```
kmadejski: As a ref: https://github.com/apache/httpd/blob/7e2d26eac309b2d79e467ef586526c10e0f226f8/include/httpd.h#L300
https://github.com/apache/httpd/blob/5f32ea94af5f1e7ea68d6fca58f0ac2478cc18c5/server/util_script.c#L495

andrerom: Thanks, added in 679a74d 👍

andrerom:
@kmadejski I think I'll: Add purging of old tags _(deprecated, but to make sure custom template usage with old style tags as was documented still works)_, in Slot code._ 

wdyt?

I'm also thinking we should actually at least warn when tags go over a certain size, so we can easier detect the cases @ITernovtsiy  (and probably also others) have issues with. We'll also need to track down the source of where his tags comes from if we end up not not handling it here.

ITernovtsiy:
I think this can be improved even more by using `c22` instead of `0c22` when no repository prefix configured. If it's not like this already)

kmadejski:
@andrerom yes, it sounds like a good idea to keep it working fine after the upgrade 🙂 

@ITernovtsiy your idea also sounds good, as for the first repository we can completely get rid of the identifier.

ITernovtsiy:
@andrerom would be good if you can make purging of old tags configurable. There are limitations to the size of that request too, so this will help there as well.

kmadejski:
@ITernovtsiy I'm not sure what you mean, but PURGE requests can be split into more than one if the tags list is too long. I believe it works like that.

ITernovtsiy:
@kmadejski right, sorry, missed it is processed in chunks already.

andrerom:
@adamwojs @alongosz 1.0 branch for 2.5LTS only ready for review, besides heavily reducing risk of reaching header limits (also see #104), this should also unblock #106.

docs/using_tags.md
```diff
@@ -127,7 +138,7 @@ Alternatively if you have a location(s) that you render inline & want invalidate
 Fo PHP usage, FOSHttpCache exposes `fos_http_cache.handler.tag_handler` service which lets you add tags to a response:
 ```php
     /** @var \FOS\HttpCache\Handler\TagHandler $tagHandler */
-    $tagHandler->addTags(['relation-33', 'relation-44']);
+    $tagHandler->addTags(['r33', 'r44']);
```
vidarl: A suggestion would be to introduce some more abstraction here so that you specify what kind of tags to add and that http-bundle based on that creates the correct key..  For instance:

`
     $tagHandler->addRelationTags([33, 44]);
`

This way we can change tag names without affecting 3rd party code ( if 3rd party code uses this abstraction instead of hard-coding the tag names

andrerom:
Closing this (PR, not branch), QA for this will be opened on meta repo.

PR #113, opened 2020-02-03
EZP-30562: As an Dev/Administrator I would like to use short HTTP cache tags

andrerom:
| Question           | Answer
| **Type**           | Bug/Improvement/Feature
| **Target version** | master
| **BC breaks**      | yes
| **Doc needed**     | yes

See: https://github.com/ezsystems/ezplatform-http-cache/pull/112

**TODO**:
- [x] Ask for Code Review.
- [x] Followup and adapt usage of tags in ezplatform-page-fieldtype: https://github.com/ezsystems/ezplatform-page-fieldtype/pull/127

### DIFF
--- a/src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
+++ b/src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
@@ -17,6 +17,7 @@ use eZ\Publish\API\Repository\Events\Content\PublishVersionEvent;
 use eZ\Publish\API\Repository\Events\Content\RevealContentEvent;
 use eZ\Publish\API\Repository\Events\Content\UpdateContentEvent;
 use eZ\Publish\API\Repository\Events\Content\UpdateContentMetadataEvent;
+use EzSystems\PlatformHttpCacheBundle\Handler\ContentTagInterface;
 
 final class ContentEventsSubscriber extends AbstractSubscriber
 {
@@ -41,9 +42,9 @@ final class ContentEventsSubscriber extends AbstractSubscriber
         $parentLocationId = $event->getDestinationLocationCreateStruct()->parentLocationId;
 
         $this->purgeClient->purge([
-            'content-' . $contentId,
-            'location-' . $parentLocationId,
-            'path-' . $parentLocationId,
+            ContentTagInterface::CONTENT_PREFIX . $contentId,
+            ContentTagInterface::LOCATION_PREFIX . $parentLocationId,
+            ContentTagInterface::PATH_PREFIX . $parentLocationId,
         ]);
     }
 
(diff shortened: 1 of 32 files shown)


ViniTou:
@mnocon 
could you look at travis build, as it fails to lacking credentials to repository (not sure why this is a case, when this is public repo).
`Installing ezsystems/doctrine-dbal-schema (dev-master 3573a95): Cloning failed using an ssh key for authentication, enter your GitHub credentials to access private repos`

andrerom:
> . It will be easier to review code, find all tag usages or change tag prefix later on, etc.

I'll open this as separate PR that we can merge across 0.9 -> 1.0 -> master as there are other concerns as well to be solved on this topic (making sure pagebuilder can use a method to get the right tags independent on changes to tags)

andrerom:
@adamwojs PR for that in https://github.com/ezsystems/ezplatform-http-cache/pull/115, please have a look asap so this can be rebase on top ;)

adamwojs:
> @adamwojs PR for that in #115, please have a look asap so this can be rebase on top ;)

@andrerom Approved!

andrerom:
@adamwojs It's been merged up here and force pushed now 👍 So ready for second look

src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
```diff
@@ -74,7 +75,7 @@ public function onDeleteVersionEvent(DeleteVersionEvent $event): void
         $contentId = $event->getVersionInfo()->getContentInfo()->id;
 
         $this->purgeClient->purge([
-            'delete-versions-' . $contentId,
+            ContentTagInterface::CONTENT_VERSION_PREFIX . $contentId,
```
andrerom: note, I assume the `delete-versions-` was a typo here and it should have been `content-version-`. If there is a new tag for our delete domain, let me know ;)

micszo:
The fix works on master (prevents FPM error) but I'm puzzled why the tags aren't shortened in browser console:

Screenshot 2020-03-03 at 13 21 30

Compared to previous result on 2.5:

Screenshot 2020-03-03 at 13 22 04

Is there a separate switch to change that?

andrerom:
Seems like you have some code using old tags? Do you have the patch for page field type applied?
_(1.0 maps old to new tags for you, master does currently not as it was deprecated in 1.0)_

micszo:
> Do you have the patch for page field type applied?

I didn't have it. I only had this branch. Now I also have the one form [130](https://github.com/ezsystems/ezplatform-page-fieldtype/pull/130).

So are long tags (at the moment) the expected behaviour on master?

andrerom:
> So are long tags (at the moment) the expected behaviour on master?

*No*, the intention is that _ALL_ should be short, but:
- you need all patches
- if there are still long tags left that means some code is still using long tags => bug

micszo:
> you need all patches

I'm on EE master with:
- ezplatform-http-cache e084ef5
- ezplatform-page-fieldtype [3247d82](https://github.com/ezsystems/ezplatform-page-fieldtype/pull/130/commits/3247d82076a2ff8118947164f2e22270708849a2)

Are these the right commits?
Is any other package crucial here?
(could the version of the VCL file have an impact?)

andrerom:
I'll test to see

> Are these the right commits?

seems so

> (could the version of the VCL file have an impact?)

no

PR #115, opened 2020-02-05
EZP-31359: Common content http cache tag abstractions

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-31359](https://jira.ez.no/browse/EZP-31359)
| **Target version** | `0.9` In order to provide future compatibility for page builder and others
| **BC breaks**      | no
| **Doc needed**     | yes, but better if this is done together with the other changes done in 1.0

Exposes abstractions and constants for the content tag prefixes which are the once documented to be used externally.

_NB: I on purpose did not use the new constant in Slots here, I'll adapt 1.0 and the corresponding master branch for that instead, less conflicts._

**TODO**:
- [x] Ask for Code Review.


andrerom:
Merging, testing will be done against 1.0/master branch and upcoming page filed type patch to use this.

PR #116, opened 2020-02-07
EZP-30562: Make it possible to ignore tags over configurable max length

andrerom:
| Question           | Answer
| **Target version** |  `1.0`
| **BC breaks**      | no
| **Doc needed**     | yes

Even if 1.0 shortens tags, it still is possible to reach the limit. On Apache (4k) and Fastly (16k) this is supposedly not configurable, so this makes it possible to cut the header, and apply a lower TTL to compensate for the risk of stale cache.

Closes #104, which proposed this for 0.9. Feature fits better with 1.0 where tags are shortened, and which is new feature release.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -28,6 +28,8 @@ class TagHandler extends FOSTagHandler implements ContentTagInterface
     private $tagsHeader;
     /** @var int|null */
     private $tagsHeaderMaxLength;
+    /** @var int|null */
+    private $tagsHeaderReducedTTl;
     /** @var LoggerInterface */
     private $logger;
 
@@ -37,13 +39,15 @@ class TagHandler extends FOSTagHandler implements ContentTagInterface
         PurgeClientInterface $purgeClient,
         RepositoryTagPrefix $prefixService,
         LoggerInterface $logger,
-        $maxTagsHeaderLength = null
+        $maxTagsHeaderLength = null,
+        $tagsHeaderReducedTTl = null
     ) {
         $this->tagsHeader = $tagsHeader;
         $this->purgeClient = $purgeClient;
         $this->prefixService = $prefixService;
         $this->logger = $logger;
         $this->tagsHeaderMaxLength = $maxTagsHeaderLength;
+        $this->tagsHeaderReducedTTl = $tagsHeaderReducedTTl;
 
         parent::__construct($cacheManager, $tagsHeader);
         $this->addTags(['ez-all']);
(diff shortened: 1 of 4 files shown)


src/Handler/TagHandler.php
```diff
@@ -88,10 +101,34 @@ static function ($tag) use ($repoPrefix) {
...
+                if (
+                    $this->tagsHeaderReducedTTl &&
+                    ($responseSharedMaxAge = $response->headers->getCacheControlDirective('s-maxage')) &&
+                    $this->tagsHeaderReducedTTl < $responseSharedMaxAge
+                ) {
+                    $response->setSharedMaxAge($this->tagsHeaderReducedTTl);
+                }
+
+                $this->logger->warning(
+                    'HTTP Cache tags header max length reached and truncated to ' . $this->tagsHeaderMaxLength
```
kmadejski: Is there any way, which wouldn't overcomplicate implementation, to enrich log message with a list of removed tags so the user can potentially manually clear these tags by himself? I'm kinda aware that it's very unlikely, but if it doesn't hurt too much it would be great to have such information in the log.

src/Handler/TagHandler.php
```diff
@@ -88,10 +101,34 @@ static function ($tag) use ($repoPrefix) {
...
+                $tagsString = substr(
+                    $tagsString,
+                    0,
+                    // Seek backwards from point of max length using negative offset
+                    strrpos($tagsString, ' ', $this->tagsHeaderMaxLength - $tagsLength)
+                );
+
+                if (
+                    $this->tagsHeaderReducedTTl &&
+                    ($responseSharedMaxAge = $response->headers->getCacheControlDirective('s-maxage')) &&
```
kmadejski: Just a small remark, would you mind moving this instruction to a separate line before `if`?

src/Handler/TagHandler.php
andrerom: This is response and not purge*, so purging those tags at this point would be pointless (they are not anymore on this response so it won'1t be cleared)

_* purge is fine, we do chunking there and several request to backend, nothing lost there_

kmadejski: Friday, 5 PM, please ignore 😄

src/Handler/TagHandler.php
andrerom: *And get rid of the micro optimization of only calling this if there is a TTL set? NEVER*

Kidding, I'll adapt it :)

src/Handler/TagHandler.php
```diff
@@ -25,16 +26,28 @@ class TagHandler extends FOSTagHandler implements ContentTagInterface
     private $purgeClient;
     private $prefixService;
     private $tagsHeader;
+    /** @var LoggerInterface */
```
mateuszbieniek: Can we use FQCN here?

src/Handler/TagHandler.php
```diff
@@ -88,10 +101,35 @@ static function ($tag) use ($repoPrefix) {
                 },
                 $tags
             );
+
             // Also add a un-prefixed `ez-all` in order to be able to purge all across repos
             $tags[] = 'ez-all';
```
mateuszbieniek: We should add this tag at the beginning of the `$tag` array or it will be the first one to go when header length will be reduced.

andrerom: Good point, fixed in b3a5d17 👍

src/Handler/TagHandler.php
andrerom: Thanks, also fixed in b3a5d17

src/Handler/TagHandler.php
```diff
@@ -25,16 +26,28 @@ class TagHandler extends FOSTagHandler implements ContentTagInterface
     private $purgeClient;
     private $prefixService;
     private $tagsHeader;
+    /** @var \Psr\Log\LoggerInterface */
+    private $logger;
+    /** @var int|null */
+    private $tagsHeaderMaxLength;
```
ViniTou: new lines between property and next doc

andrerom:
Merged in https://github.com/ezsystems/ezplatform-http-cache/commit/c94402b9c67f2fed2dc64530ee59a9465eaf3d33

PR #117, opened 2020-02-10
EZP-31374: Cannot autowire service that has a dependency on TagHandler

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-31374](https://jira.ez.no/browse/EZP-31374)
| **Type**           | Bug
| **Target version** | `0.9`
| **BC breaks**      | no
| **Doc needed**     | no

For more info please refer to JIRA ticket. 

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/DependencyInjection/Compiler/DriverPass.php
+++ b/src/DependencyInjection/Compiler/DriverPass.php
@@ -8,6 +8,7 @@ namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use EzSystems\PlatformHttpCacheBundle\Handler\TagHandler;
 
 /**
  * We support http cache drivers to be provided by 3rd party bundles.
@@ -35,7 +36,7 @@ class DriverPass implements CompilerPassInterface
         }
 
         $container->setAlias('fos_http_cache.handler.tag_handler', $configuredFosTagHandlerServiceId);
-        $container->setAlias('EzSystems\PlatformHttpCacheBundle\Handler\TagHandler', $configuredFosTagHandlerServiceId);
+        $container->setAlias(TagHandler::class, $configuredFosTagHandlerServiceId);
     }
 
     public static function getTaggedService(ContainerBuilder $container, $tag)
--- a/src/DependencyInjection/Compiler/DriverPass.php
+++ b/src/DependencyInjection/Compiler/DriverPass.php
@@ -35,6 +35,7 @@ class DriverPass implements CompilerPassInterface
         }
 
         $container->setAlias('fos_http_cache.handler.tag_handler', $configuredFosTagHandlerServiceId);
+        $container->setAlias('EzSystems\PlatformHttpCacheBundle\Handler\TagHandler', $configuredFosTagHandlerServiceId);
     }
 
     public static function getTaggedService(ContainerBuilder $container, $tag)


webhdx:
Why are you aliasing existing class so the service can point to completely different class instead of making an alias on an interface?

alongosz:
#### TL;DR;
I agree with @webhdx 

#### Reason:

TBH compiler pass shouldn't be involved when choosing tag handler based on a Container parameter. It's a responsibility of either a _Registry_ or a _Factory_. There should be a common interface for anything which is injectable. Interface should be built by a _Factory_ (_Registry_ should be injected into _Factory_, but that might be over-engineered a bit).

The minimum here is identifying a common SPI interface and use it for auto-wiring. Alternatively dynamic decoration could be used, but seems to me it might be too complicated for the simple case. Aliasing everything during compile time is error-prone and difficult to debug (hence the issue at hand).

kmadejski:
@webhdx & @alongosz thanks for the feedback.

> Why are you aliasing existing class so the service can point to completely different class instead of making an alias on an interface?

What interface you have in mind in particular? Did you think about this one: https://github.com/ezsystems/ezplatform-http-cache/blob/0.9/src/Handler/TagHandlerInterface.php? It's deprecated for a quite long time and never used anywhere. 

I'm aliasing this class because there is a couple of external classes that depend on *on class* not interface, for instance: https://github.com/ezsystems/ezplatform-page-fieldtype/blob/1.3/src/lib/Event/Subscriber/TagBlockResponseSubscriber.php#L55

> There should be a common interface for anything which is injectable. Interface should be built by a Factory (Registry should be injected into Factory, but that might be over-engineered a bit).
The minimum here is identifying a common SPI interface and use it for auto-wiring.

I definitely agree, but that simply means changes in a number of different repositories. I simply wanted to unblock for the current LTS. In the packages for v3, it would be more than great to have such an interface.

andrerom:
> I'm aliasing this class because there is a couple of external classes that depend on class not interface, for instance: https://github.com/ezsystems/ezplatform-page-fieldtype/blob/master/src/lib/FieldType/Page/Block/Event/Listener/ContentListBlockListener.php

So then you need this for master, and not so much 2.5 then.

Be aware the following PR for page-field-type: https://github.com/ezsystems/ezplatform-page-fieldtype/pull/127
Will imply the code you point to can ALMOST type hint with the new `EzSystems\PlatformHttpCacheBundle\Handler\ContentTagInterface` instead

Obstacle is this: `$this->tagHandler->tagSymfonyResponse($response);`
We could make `ContentTagInterface` extend the interface from FOS to avoid that being an issue.

webhdx:
Too bad there is no common interface for tag handlers. I'd say we should fix all places where `TagHandler` is autowired and switch to direct `@fos_http_cache.handler.tag_handler` injection.

lserwatka:
I agree with @webhdx as this looks like a quick fix but not a proper problem resolution.

alongosz:
> What interface you have in mind in particular? Did you think about this one: https://github.com/ezsystems/ezplatform-http-cache/blob/0.9/src/Handler/TagHandlerInterface.php? It's deprecated for a quite long time and never used anywhere.

I'm a bit lost. We [document that TagHandler needs to implement `TagHandlerInterface`](https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/drivers.md#taghandler), but that interface is deprecated?

This could lead to some issues in the future. You cannot simply swap one concrete implementation with another in hope it will work. The doc also does not state that auto-wiring is supported, **so the fix would be rather to inject concrete implementation in those specific places.**

Anything that is injectable should have its own corresponding interface providing contracts to be used.

_Nitpick: Please do not keep documentation in `/doc` folders. Everything that is supposed to be official should be placed in [`ezsystems/developer-documentation`](https://github.com/ezsystems/developer-documentation)._

andrerom:
> I'm a bit lost. We document that TagHandler needs to implement TagHandlerInterface, but that interface is deprecated?

ignore `docs/drivers.md`, it's highly outdated and will need several updates as followup to this.

kmadejski:
@webhdx / @alongosz as said before, I agree with you guys that it should be done properly by introducing a new, proper interface. The question is whether we want to do it for v2.5 or rather v3?
My goal here was to unblock it for LTS, I'm more than happy to see it done properly for v3.

webhdx:
Sure, have you considered fixing all places to direct `@fos_http_cache.handler.tag_handler` injection? I'd say it's a bug in all these places as `TagHandler` is a concrete implementation and shouldn't be used for autowiring unless you are compositing or decorating it.

kmadejski:
I'm closing this one in favor of: https://github.com/ezsystems/ezplatform-page-fieldtype/pull/128

PR #118, opened 2020-02-18
[3.x][VCL] Update purge logic from FOS

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-XXXXX](https://jira.ez.no/browse/EZP-XXXXX)
| **BC breaks**      | 🤷‍♂ 
| **Doc needed**     | maybe

Updates VCL to be in sync with the purge logic from FOS.
This drops eZ specific Ban which was Varnish specific, and adds support for FOS capability to do hard tag purge.

### DIFF
--- a/docs/varnish/vcl/varnish5.vcl
+++ b/docs/varnish/vcl/varnish5.vcl
@@ -127,32 +127,33 @@ sub ez_purge {
     // Retrieve purge token, needs to be here due to restart, match for PURGE method done within
     call ez_invalidate_token;
 
-    # Support how purging was done in earlier versions, this is deprecated and here just for BC for code still using it
-    if (req.method == "BAN") {
+    # Adapted with acl from vendor/friendsofsymfony/http-cache/resources/config/varnish/fos_tags_xkey.vcl
+    if (req.method == "PURGEKEYS") {
         call ez_purge_acl;
 
-        if (req.http.X-Location-Id) {
-            ban("obj.http.X-Location-Id ~ " + req.http.X-Location-Id);
-            if (client.ip ~ debuggers) {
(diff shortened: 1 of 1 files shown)


docs/varnish/vcl/varnish5.vcl
```diff
@@ -127,32 +127,33 @@ sub ez_purge {
     // Retrieve purge token, needs to be here due to restart, match for PURGE method done within
     call ez_invalidate_token;
```
andrerom: Not a regression from this PR, however the code in this subroutine will need to be adapted to also take into account `req.method == "PURGEKEYS"`.

Ref: `if (req.restarts == 0 && req.method == "PURGE" && req.http.x-invalidate-token) {`

andrerom: Proposal for fix on that in e377e19

micszo:
@andrerom What is the recommended way of switching between `purge` and `softpurge` in this approach? 🙂

andrerom:
@micszo It's a FOSHttpCache config: https://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/proxy-client.html#tag-mode

micszo:
Nota bene, @mnocon aptly noticed that there also is https://github.com/ezsystems/ezplatform/blob/master/.platform/varnish.vcl. Will that one have to be adapted to these changes as well?

andrerom:
> Will that one have to be adapted to these changes as well?

not by me, I'd suggest we find a way to drop it and use the one from here.

Same with this btw:
https://github.com/ezsystems/ezplatform/blob/master/doc/varnish/vcl/varnish4_xkey.vcl

andrerom:
@micszo Changes for meta repo: https://github.com/ezsystems/ezplatform/pull/505

PR #119, opened 2020-02-19
EZP-30560: Unable to clear Varnish cache on Platform.sh

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-30560](https://jira.ez.no/browse/EZP-30560)
| **Type**           | Bug
| **Target version** | `0.9`
| **BC breaks**      | no
| **Doc needed**     | no

Let's say there is eZ Platform site on Platform.sh. Own domain is used to access the admin. So site has following domains:
- https://project.com for front-end
- https://admin.project.com for admin UI

Front-end (https://project.com) is served by Varnish, but admin is not:
```
// .platform/routes.yaml
https://project.com:
    type: upstream
    upstream: varnish:http
https://admin.project.com:
    type: upstream
    upstream: app:http
```

When you edit/publish some content in admin UI (https://admin.project.com), it is expected that its Varnish cache will be cleared. But actually, the Varnish cache is not cleared.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/PurgeClient/VarnishPurgeClient.php
+++ b/src/PurgeClient/VarnishPurgeClient.php
@@ -85,7 +85,7 @@ class VarnishPurgeClient implements PurgeClientInterface
     private function getPurgeHeaders()
     {
         $host = parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST);
-        $fallbackHost = $_SERVER['SERVER_NAME'] ?? 'localhost';
+        $fallbackHost = empty($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME']) : 'localhost';
 
         $headers = [
             'Host' => $host ?: $fallbackHost,
--- a/src/PurgeClient/VarnishPurgeClient.php
+++ b/src/PurgeClient/VarnishPurgeClient.php
@@ -84,8 +84,11 @@ class VarnishPurgeClient implements PurgeClientInterface
      */
     private function getPurgeHeaders()
     {
+        $host = parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST);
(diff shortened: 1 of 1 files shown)


src/PurgeClient/VarnishPurgeClient.php
```diff
@@ -84,8 +84,10 @@ public function purgeAll()
      */
     private function getPurgeHeaders()
     {
+        $host = parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST);
+
         $headers = [
-            'Host' => empty($_SERVER['SERVER_NAME']) ? parse_url($this->configResolver->getParameter('http_cache.purge_servers')[0], PHP_URL_HOST) : $_SERVER['SERVER_NAME'],
+            'Host' => $host ?: $_SERVER['SERVER_NAME'],
```
andrerom: Looks good, only think that comes to mind is PHP notice when this fallback server var is sometimes not set, suggestions?

SerheyDolgushev: Can you please check https://github.com/ezsystems/ezplatform-http-cache/pull/119/commits/f63f51e8a351559a34f15136584bef3f173f02bf?

andrerom: I think you misunderstood me, my point was about ` $_SERVER['SERVER_NAME']` now being used w/o check if it is set first.

To avoid having to adapt tests I propose you revert f63f51e
And that should be ok, my comment can be ignored, there is anyway nothing to fallback to form SERVER_NAME unless we hard code a last fallback to `localhost` like we did [before](https://github.com/ezsystems/ezplatform-http-cache/blob/v0.8.0/src/PurgeClient/VarnishPurgeClient.php#L46).

lserwatka:
@adamwojs could you merge it up?

PR #120, opened 2020-03-17
EZP-31493: Removed http as valid purge type

ViniTou:
| Question           | Answer
| **JIRA issue**     | [EZP-31493](https://jira.ez.no/browse/EZP-31493)
| **Type**           | Misc
| **Target version** | `master` for new features
| **BC breaks**      | yes
| **Doc needed**     | yes

It seems that based on platform.sh configuration only `varnish` is proper purge_type, so it seems to be better to remove one and stick with `varnish` instead of `http` as it also seems exacly what Fos did.

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/DependencyInjection/EzPlatformHttpCacheExtension.php
+++ b/src/DependencyInjection/EzPlatformHttpCacheExtension.php
@@ -37,9 +37,6 @@ class EzPlatformHttpCacheExtension extends Extension implements PrependExtension
         $loader->load('services.yml');
         $loader->load('event.yml');
         $loader->load('view_cache.yml');
-
-        $this->setDefaultResponseHeader($container);
-        $this->setDefaultTagSeparator($container);
     }
 
     public function prepend(ContainerBuilder $container)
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/EzPlatformHttpCacheExtension.php
```diff
@@ -69,28 +66,4 @@ public function getExtraConfigParsers()
...
     }
-
-    /**
-     * Overrides default header name based on setting tag_mode in FosHttpCacheBundle configuration.
-     */
-    private function setDefaultResponseHeader(ContainerBuilder $container): void
-    {
-        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
-
-        $responseHeader = 'http' === $purgeType ? 'xkey' : TagHeaderFormatter::DEFAULT_HEADER_NAME;
```
andrerom: Side note: This probably need adoption in 0.9 and up as well then after @vidarl introduced `varnish` as alias to me more clear on varnish vs fastly.

andrerom: Ignore, this is master only logic.

Good that you remove it 👍

alongosz:
Bad news Guys, had to revert most of this PR via #124. For `local` purge type it's still needed. Spent hours debugging REST failure... ;)

PR #123, opened 2020-03-20
EZP-31505: Varnish config is loaded only when varnish set as purge_type

ViniTou:
| Question           | Answer
| **JIRA issue**     | [EZP-31505](https://jira.ez.no/browse/EZP-31505)
| **Type**           | Bug
| **Target version** |  `master`
| **Doc needed**     | no

Instead of changing default variables that are set by fos when varnish config is used, that yaml file is only included when `varnish` is set as `purge_type`.
That involved change in order of bundle loading:
https://github.com/ezsystems/ezplatform/pull/512

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/DependencyInjection/Compiler/VarnishCachePass.php
+++ b/src/DependencyInjection/Compiler/VarnishCachePass.php
@@ -6,23 +6,26 @@
  */
 namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection\Compiler;
 
+use EzSystems\PlatformHttpCacheBundle\ProxyClient\Varnish;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
 
 class VarnishCachePass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
+        if (!$container->hasDefinition('fos_http_cache.proxy_client.varnish')) {
+            $container->removeDefinition('ezplatform.http_cache.proxy_client.varnish.http_dispatcher');
+            return;
+        }
+
+        $this->overrideDefaultProxyClient($container);
         $this->processVarnishProxyClientSettings($container);
     }
 
-    private function processVarnishProxyClientSettings(ContainerBuilder $container)
+    private function processVarnishProxyClientSettings(ContainerBuilder $container): void
     {
-        if (!$container->hasDefinition('fos_http_cache.proxy_client.varnish')) {
-            throw new InvalidArgumentException('Varnish proxy client must be enabled in FOSHttpCacheBundle');
-        }
-
         $fosConfig = array_merge(...$container->getExtensionConfig('fos_http_cache'));
 
         $servers = $fosConfig['proxy_client']['varnish']['http']['servers'] ?? [];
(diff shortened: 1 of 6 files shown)


src/DependencyInjection/Compiler/VarnishCachePass.php
```diff
@@ -38,4 +41,15 @@ private function processVarnishProxyClientSettings(ContainerBuilder $container)
...
+
+    private function overrideDefaultProxyClient(ContainerBuilder $container): void
+    {
+        $varnishProxyClient = $container->getDefinition('fos_http_cache.proxy_client.varnish');
+        $varnishProxyClient->setClass(Varnish::class);
+        $varnishProxyClient->setArguments([
+            new Reference('ezpublish.config.resolver'),
+            new Reference('fos_http_cache.proxy_client.varnish.http_dispatcher'),
+            $container->getParameter('fos_http_cache.proxy_client.varnish.options'),
+        ]);
```
alongosz: Could you decorate (via `decorates` attribute in Yaml) instead of overriding?

src/ProxyClient/Varnish.php
```diff
@@ -35,8 +35,9 @@ public function __construct(
 
     private function fetchAndMergeAuthHeaders(array $headers): array
     {
-        if ($this->configResolver->hasParameter('http_cache.varnish_invalidate_token')) {
-            $headers[InvalidateTokenController::TOKEN_HEADER_NAME] = $this->configResolver->getParameter('http_cache.varnish_invalidate_token');
+        if ($this->configResolver->hasParameter('http_cache.varnish_invalidate_token') &&
+            '' !== $token = $this->configResolver->getParameter('http_cache.varnish_invalidate_token')) {
```
alongosz: CS: this is not readable to me
```suggestion
        if (
            $this->configResolver->hasParameter('http_cache.varnish_invalidate_token')
            && '' !== ($token = $this->configResolver->getParameter(
                'http_cache.varnish_invalidate_token'
            ))
        ) {
```

src/DependencyInjection/Compiler/VarnishCachePass.php
ViniTou: I tried, but that class is very convoluted with Abstract.

alongosz: Ok, let's not spend too much time on this now. Thanks for trying :+1:

src/ProxyClient/Varnish.php
```diff
@@ -36,7 +36,10 @@ public function __construct(
     private function fetchAndMergeAuthHeaders(array $headers): array
     {
         if ($this->configResolver->hasParameter('http_cache.varnish_invalidate_token')) {
-            $headers[InvalidateTokenController::TOKEN_HEADER_NAME] = $this->configResolver->getParameter('http_cache.varnish_invalidate_token');
+            $token = $this->configResolver->getParameter('http_cache.varnish_invalidate_token');
+            if ('' !== $token) {
+                $headers[InvalidateTokenController::TOKEN_HEADER_NAME] = $token;
```
andrerom:
```suggestion
            $purgeAuthToken = $this->configResolver->getParameter('http_cache.varnish_invalidate_token');
            if ('' !== $purgeAuthToken) {
                $headers[InvalidateTokenController::TOKEN_HEADER_NAME] = $purgeAuthToken;
```

Or something similar to make it more clear what it is.

PR #124, opened 2020-03-21
Restored FOS header and separator override for local purge type

alongosz:
| Question           | Answer
| **JIRA issue**     | Follow up to [EZP-31493](https://jira.ez.no/browse/EZP-31493) (#120)
| **Type**           | Bug
| **BC breaks**      | no
| **Doc needed**     | no

### Summary

While `http` purge type is no longer needed for Varnish, #120 also removed overriding FOS Sf Container parameters for `local` purge type. It is still default purge type available for any environment other than dev. The following FOSHttpCache parameters were restored:
- `fos_http_cache.tag_handler.response_header`
- `fos_http_cache.tag_handler.separator`

I've discovered this issue while debugging [REST failure](https://travis-ci.org/github/ezsystems/ezplatform-rest/jobs/664966144#L1305). Please observe that when meta is set with the branch from this PR it [no longer fails](https://travis-ci.org/github/ezsystems/ezplatform-rest/jobs/665053373).

### Further issues

I'm not that familiar with HTTP cache, so it was a bit of trial and error attempt to unblock entire CI. Unfortunately this means that ezsystems/ezplatform#512 will create this regression once again because FOS will override those settings.

I have a feeling the approach here is not the best and can be configured differently, but no idea how (TL;DR; FOSHttpCache doc). Unless it no longer supports anything other than Varnish?

Alternatively maybe instead of using `X-Cache-Tags` (`TagHeaderFormatter::DEFAULT_HEADER_NAME`) we could use `xkey` and space as separator for `local` as well? If this is nonsense, forgive me :-)

**TODO**:
- [x] Restore FOSHttpCache parameters override which cause local purge type to fail.
- [x] Wait for Travis (at least for jobs which should pass now).
- [ ] Ask for Code Review.

### DIFF
--- a/src/DependencyInjection/EzPlatformHttpCacheExtension.php
+++ b/src/DependencyInjection/EzPlatformHttpCacheExtension.php
@@ -7,6 +7,7 @@
 namespace EzSystems\PlatformHttpCacheBundle\DependencyInjection;
 
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ParserInterface;
+use FOS\HttpCache\TagHeaderFormatter\TagHeaderFormatter;
 use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Config\FileLocator;
@@ -36,6 +37,15 @@ class EzPlatformHttpCacheExtension extends Extension implements PrependExtension
         $loader->load('services.yml');
         $loader->load('event.yml');
         $loader->load('view_cache.yml');
+
+        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+        if ('local' === $purgeType) {
+            $container->setParameter(
+                'fos_http_cache.tag_handler.response_header',
+                TagHeaderFormatter::DEFAULT_HEADER_NAME
+            );
+            $container->setParameter('fos_http_cache.tag_handler.separator', ',');
+        }
     }
 
     public function prepend(ContainerBuilder $container)


src/DependencyInjection/EzPlatformHttpCacheExtension.php
```diff
@@ -36,6 +37,15 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.yml');
         $loader->load('event.yml');
         $loader->load('view_cache.yml');
+
+        $purgeType = $container->getParameter('ezpublish.http_cache.purge_type');
+        if ('local' === $purgeType) {
```
alongosz: We had `http` which is gone, have `varnish` and this one. Was there anything else?

alongosz:
FYI @ViniTou @andrerom we didn't detect this issue because CI was failing on dependency issue.

All is green now, so I need to merge it as-is to unblock meta, but of course the questions here are still valid and if you have an alternative solution, I will make a follow-up.

ViniTou:
@alongosz 
It was detected, and I said it in daily, and there is also PR to fix it better way.

ViniTou:
As for used local storage implementation, afair headers could be changed from default, but separator is hardcoded inside library.

alongosz:
> @alongosz
> It was detected, and I said it in daily, and there is also PR to fix it better way.

@ViniTou hmm, first thing I did once I figured out http-cache is responsible, was to checkout #123 and its corresponding ezsystems/ezplatform#512. Either I made a mistake somewhere during checkout or the issue was still there. Unless you're talking about different PR ;)

src/DependencyInjection/EzPlatformHttpCacheExtension.php
andrerom: `fastly`, but in other EE bundle

PR #133, opened 2020-07-29
Fix Adding X-Invalidate-Token header only when TOKEN is not null

kaff:
| Question           | Answer
| **JIRA issue**     | [EZP-31607](https://jira.ez.no/browse/EZP-31607)
| **Type**           | Bug
| **Target version** | `2.0`
| **BC breaks**      | no
| **Doc needed**     | no

Prevent attaching X-Invalidate-Token header to Varnish request when http_cache.varnish_invalidate_token parameter is not set or null.

**TODO**:
- [x] Implement fix a bug.
- [x] Ask for Code Review.

### DIFF
--- a/spec/ProxyClient/VarnishSpec.php
+++ b/spec/ProxyClient/VarnishSpec.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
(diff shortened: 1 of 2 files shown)


ViniTou:
@andrerom 
If I am right this seems to be ok in `1.0`.

https://github.com/ezsystems/ezplatform-http-cache/blob/1.0/src/PurgeClient/VarnishPurgeClient.php#L94

andrerom:
> If I am right this seems to be ok in 1.0.

you are very much right :)

PR #135, opened 2020-10-01
Fix discrepancy between doc and twig template name

andrerom:
| Question           | Answer
| **JIRA issue**     | EZP-XXXXX
| **Type**           | Bug
| **Target version** | 2.1+
| **BC breaks**      | no, rather a BC fix depending on how you look at it
| **Doc needed**     | no, doc is already correct if this is merged

Fixing name break in 3.0, as both doc, and the other methods have different syntax by now I suggest we simply close this in the following way, and cleanup when moving to new prefixes at some point in the future. _IMO maybe with a bit better BC next time though._

Reported in: https://github.com/ezsystems/developer-documentation/pull/1132#discussion_r497281925

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/Twig/ContentTaggingExtension.php
+++ b/src/Twig/ContentTaggingExtension.php
@@ -40,6 +40,11 @@ class ContentTaggingExtension extends AbstractExtension
                 'ez_http_cache_tag_location',
                 [$this, 'tagHttpCacheForLocation']
             ),
+            // For 2.5 BC, and to be consistent with the other functions, to be cleaned up with new prefix in the future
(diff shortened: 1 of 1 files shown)


adamwojs (approved):

vidarl:
ah.. shoot. I was 1 min too late..
I would suggest we call all functions `ez_http_cache`, but that we also keep the `ez_http` function names but mark them deprecated...

ViniTou:
I have merged it up together with https://github.com/ezsystems/ezplatform-http-cache/pull/137

PR #136, opened 2020-10-09
EZP-32017: fix trusted proxy setup for platform.sh

ITernovtsii:
| Question           | Answer
| **JIRA issue**     | [EZP-32017](https://jira.ez.no/browse/EZP-32017)
| **Type**           | Bug
| **Target version** | `1.0`
| **BC breaks**      | no
| **Doc needed**     | no

platform.sh use "ngx_http_realip_module" nginx module, and put client IP into REMOTE_ADDR.
This makes it impossible to configure an application to be accessible from both with and without varnish.

If you'll put REMOTE_ADDR in TRUSTED_PROXIES, admin panel will be open to IP address spoof using "x-forwarded-for" header.
And, the user will be able to request your varnish invalidation token.

If you'll put other IP addresses, or don't specify TRUSTED_PROXIES at all, you'll receive "Unauthorized" from varnish (because of "/_ez_http_invalidatetoken" route checking "$request->isFromTrustedProxy()", which will return false).

This PR makes it possible to have varnish without defined TRUSTED_PROXIES

Related to https://github.com/ezsystems/ezplatform/pull/609 but completely different)

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/Controller/InvalidateTokenController.php
+++ b/src/Controller/InvalidateTokenController.php
@@ -14,6 +14,8 @@ use eZ\Publish\Core\MVC\ConfigResolverInterface;
 
 class InvalidateTokenController
 {
+    public const TOKEN_HEADER_NAME = 'X-Invalidate-Token';
+
     /**
      * @var \eZ\Publish\Core\MVC\ConfigResolverInterface
      */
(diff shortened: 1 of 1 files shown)


ITernovtsiy:
let me know if it's not something you would like to release for 2.5 and I'll rebase to 3.2

src/Controller/InvalidateTokenController.php
```diff
@@ -66,25 +68,27 @@ public function tokenAction(Request $request)
...
             return $response;
         }
 
-        // Important to keep this condition, as .vcl rely on this to prevent everyone from being able to fetch the token.
-        if ($request->headers->get('accept') !== 'application/vnd.ezplatform.invalidate-token') {
-            $response->setStatusCode('400', 'Bad request');
+        $this->tagHandler->addTags(['ez-invalidate-token']);
+
+        $token = $this->configResolver->getParameter('http_cache.varnish_invalidate_token');
+        if ($request->headers->get(self::TOKEN_HEADER_NAME) !== $token) {
```
vidarl, line 81: I don't get this...
so, the trusted proxy varnish does not know the invalidate token, so it will try to [retrieve it](https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/varnish/vcl/varnish5.vcl#L234) by accessing `/_ez_http_invalidatetoken`. But now that is rejected because it doesn't have the invalidation token....

Or am I missing something?

vidarl (changes requested):

src/Controller/InvalidateTokenController.php
vidarl: ah... It will use the token it received in the initial PURGE request

ITernovtsiy: Yes, it's my favorite part of this PR)

ITernovtsiy: done

vidarl (changes requested):
So, the purpose of `tokenAction()` is no longer to actually return the invalidate-token, but to validate that a given token is correct.

To make that clear to everybody, I think we should state that in PHP doc for this method. 

To better reflect this change, the controller should maybe no longer then return the token? Instead, the .vcl could check http status in the response. That will cause a BC though, (Things will breake if someone would update bundle, but not .vcl).
So, maybe just leave the logic in .vcl as is. I'll let others comment on that

ITernovtsiy:
> So, the purpose of tokenAction() is no longer to actually return the invalidate-token, but to validate that a given token is correct.
> To make that clear to everybody, I think we should state that in PHP doc for this method.

I would like to keep the token returned to varnish. Varnish will cache it and don't bother PHP for PURGE requests with invalid token. But this will work only if the cache doesn't Vary by `X-Invalidate-Token`.
I think it doesn't, but I'll recheck to be sure.

ITernovtsiy:
@vidarl I rechecked and see token validated properly (on varnish) for requests with invalid token, without sending `/_ez_http_invalidatetoken` to backend. So, I would vote to keep PR as is, without BC changes.

vidarl (approved):
Ok. Fine by me. I'll leave it up to @ezsystems/php-dev-team to reflect a bit on my comment: https://github.com/ezsystems/ezplatform-http-cache/pull/136#pullrequestreview-553472728

PR #137, opened 2020-10-09
EZP-32020: make /_ez_http_invalidatetoken cached by varnish

ITernovtsiy:
| Question           | Answer
| **JIRA issue**     | [EZP-32020](https://jira.ez.no/browse/EZP-32020)
| **Type**           | Bug
| **Target version** | `2.1`+
| **BC breaks**      | no
| **Doc needed**     | no

Currently, each request to purge cache by tags triggers a request to "/_ez_http_invalidatetoken" route
A response of this route should be cached by varnish, but it's not.
It's caused by session usage, and the fact this response does not vary by user context.
So, Symfony listener changing response to private.

PS. the same `NO_AUTO_CACHE_CONTROL_HEADER` used in fos symfony cache bundle

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/Controller/InvalidateTokenController.php
+++ b/src/Controller/InvalidateTokenController.php
@@ -10,6 +10,7 @@ use FOS\HttpCache\ResponseTagger;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Request;
 use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use Symfony\Component\HttpKernel\EventListener\SessionListener;
 
 class InvalidateTokenController
 {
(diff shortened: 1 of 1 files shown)


andrerom (approved):

andrerom:
should be applied to 1.0 as well or is Symfony 3.4 perhaps not affected? (the FosHrrpCache change you refer to might give a hint on that)

ITernovtsiy:
Symfony 3.4 is not affected.

There is a version check in FOSHttpCacheBundle:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/EventListener/UserContextListener.php#L167

PS. failed tests are not related to PR changes.

vidarl (approved):

ViniTou (approved):

andrerom:
I realize we probably need to make sure this is always cleared on deploy hook with p.sh, as the invalidation token there based on `PLATFORM_PROJECT_ENTROPY` which changes on every deploy.

vidarl:
`PLATFORM_PROJECT_ENTROPY ` should remain the same for the whole project lifetime and not change on deploy: https://docs.platform.sh/development/variables.html#platformsh-provided-variables

andrerom:
True, I was thinking of `PLATFORM_TREE_ID `, but then we are good here still.

Steveb-p (approved):

mnocon (approved):
Tested on eZ Platform 3.1 that the token is cached and next requests to purge cache do not hit the endpoint providing the token.

Approved!

lserwatka:
@ViniTou could you merge it up?

PR #138, opened 2020-11-06
EZP-32103: Add Varnish integration tests

mnocon:
| Question           | Answer
| **JIRA issue**     | [EZP-32103](https://issues.ibexa.co/browse/EZP-32103)
| **Type**           | Tests
| **Target version** | 2.1,2.2 and master
| **BC breaks**      | no
| **Doc needed**     | no

Requires: 
1) https://github.com/ezsystems/ezplatform/pull/621
2) https://github.com/ezsystems/BehatBundle/pull/166

Things done:
1) Fixed the test for Symfony cache meant for admin user - without password it was running as anonymous 😅 
2) Added two jobs running tests with Varnish - one with and one without invalidate token
3) Added basic tests for Varnish - making sure that pages are cached for anonymous and logged in users, also that the cache is refreshed correctly on Content changes.

**TODO**:
- [x] Ask for Code Review.
- [x] **Rebase before merging - remove TMP commit**

### DIFF
--- a/features/varnish/cache.feature
+++ b/features/varnish/cache.feature
@@ -1,0 +1,60 @@
+@varnish
+Feature: As an site administrator I want my pages to be cached using Varnish
+
+    @admin
+    Scenario Outline: Content Items are cached for user when visited
+        Given I create "Folder" Content items in root in "eng-GB"
+            | name       | short_name |
+            | TestFolder | <itemName> |
+        And I am viewing the pages on siteaccess "site" as "<user>"
+        When I visit <itemName> on siteaccess "site"
+        And response headers contain
+            | Header  | Value |
+            | X-Cache | MISS  |
+        And I reload the page
+        Then I see correct preview data for "Folder" Content Type
+            | field | value      |
+            | title | <itemName> |
+        And response headers contain
+            | Header  | Value |
+            | X-Cache | HIT   |
+
+        Examples:
+            | user      | password | itemName                     |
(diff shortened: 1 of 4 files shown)


features/varnish/cache.feature
```diff
@@ -0,0 +1,60 @@
...
+            | field | value      |
+            | title | <itemName> |
+        And response headers contain
+            | Header  | Value |
+            | X-Cache | HIT   |
+        When I edit "<itemName>" Content item in "eng-GB"
+            | short_name          |
+            | <itemNameAfterEdit> |
+        And I reload the page
+        # Give Varnish time to fetch the backend response
```
mnocon, line 45: This is tricky: if I understand correctly, on Content edit the cache is soft-purged - meaning that Varnish will keep serving the old cache until it it fetched from the backend again.

The first reload triggers the backend fetch, but if we follow immediately with another reload we will still be served the stale content - we need to wait until the fetch is completed.

On my local env 1 second was enough for the test to pass, I've chosen 5 for CI safety.

features/varnish/cache.feature
```diff
@@ -0,0 +1,60 @@
...
+        And I wait 5 seconds
+        # Second reload is needed because of soft purging
+        And I reload the page
+        And I reload the page
```
mnocon, line 49: Without this (third) reload the response would still be marked with "X-Cache: miss". Not sure why TBH.

micszo (approved):
👍

bogusez (approved):

PR #166, opened 2022-01-03
IBX-1784: Introduced translation-aware HTTP cache invalidation on content publication

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | [IBX-1784](https://issues.ibexa.co/browse/IBX-1784)
| **Target version** | `2.3`
| **BC breaks**      | no
| **Doc needed**     | yes

Requires: https://github.com/ezsystems/ezplatform-kernel/pull/279
Fastly changes: https://github.com/ezsystems/ezplatform-http-cache-fastly/pull/35
Recipes: https://github.com/ibexa/recipes/pull/72

This PR introduces changes to HTTP Cache purging on a version published which improves performance by allowing to purge only HTTP Cache relevant for translation's language. Currently, when any translation is published, all HTTP cache related to the content is purged, causing for the HTTP cache not affected by the content change to be rebuilt unnecessarily.

This PR introduces some changes in VCL, tagging, and purging.

**Tagging:**
Before sending the content's response, a new `xlang` header containing the translation's language code is added. When `prioritizedFieldLanguageCode` is `null` it means that the Content requested is a fallback for missing translation, therefore  `initialLanguageCode` is used to avoid purging issues.

In VCL, `xkey` header is duplicated, modified by adding `xlang` value as a suffix to the duplicated tags, and used to tag response with original `xkey`. Additional tags have to be introduced for proper invalidation but to avoid issues with the webserver's maximum header size, they are synthesized inside VCL.

**Purging:**
When a new version is published, all tags are tagged with a translation's Language Code suffix and invalidated (we don't have to worry about maximum header size here, as the request is chunked by `ProxyClient`). 
This occurs when the translation is not the main translation or when the content's Content Type does not have non-translatable fields.

If those conditions are not met, HTTP cache is purged without language code suffix, invalidating all content-related HTTP cache.

**Potential problems:**
I've planed to add `xlang` header to the purge requests as well - which would be a more elegant way to handle tags with language code suffix inside VCL (there is no risk of cache purging issues when someone upgrades packages version, but won't modify the VCL). Unfortunately, I wasn't able to find a way to do that without ugly hacking. Maybe someone has an idea how/if it can be done?

**TODO**:
- [x] Add changes to Fastly's vcl
- [x] Add tag for clearing all translations from template
- [x] Invalidate AA content when translation is created 
- [x] Ask for Code Review.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\PlatformHttpCacheBundle\\": "src",
-            "EzSystems\\PlatformHttpCacheBundle\\Tests\\": "tests"
+            "EzSystems\\PlatformHttpCacheBundle\\Tests\\": "tests",
+            "Ibexa\\HttpCache\\": "src"
         }
     },
     "scripts": {
--- a/src/EventSubscriber/AddContentLanguageHeaderSubscriber.php
+++ b/src/EventSubscriber/AddContentLanguageHeaderSubscriber.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\HttpCache\EventSubscriber;
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\Core\MVC\Symfony\View\CachableView;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+final class AddContentLanguageHeaderSubscriber implements EventSubscriberInterface
+{
+    public const CONTENT_LANGUAGE_HEADER = 'x-lang';
+
+    /** @var bool */
+    private $isTranslationAware;
+
+    public function __construct(bool $isTranslationAware)
+    {
+        $this->isTranslationAware = $isTranslationAware;
+    }
+
+    public function onKernelResponse(ResponseEvent $event)
+    {
+        if (!$this->isTranslationAware || HttpKernelInterface::MASTER_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        $view = $request->attributes->get('view');
+        if (!$view instanceof CachableView || !$view->isCacheEnabled()) {
+            return;
+        }
+
+        $content = $request->attributes->get('content');
+        if ($content instanceof Content) {
+            $event->getResponse()->headers->add([self::CONTENT_LANGUAGE_HEADER => $content->getDefaultLanguageCode()]);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::RESPONSE => 'onKernelResponse',
+        ];
+    }
+}
--- a/src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
+++ b/src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
@@ -17,10 +17,36 @@ use eZ\Publish\API\Repository\Events\Content\PublishVersionEvent;
 use eZ\Publish\API\Repository\Events\Content\RevealContentEvent;
 use eZ\Publish\API\Repository\Events\Content\UpdateContentEvent;
 use eZ\Publish\API\Repository\Events\Content\UpdateContentMetadataEvent;
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
+use eZ\Publish\SPI\Persistence\Content\Handler as ContentHandler;
+use eZ\Publish\SPI\Persistence\Content\Location\Handler as LocationHandler;
+use eZ\Publish\SPI\Persistence\URL\Handler as UrlHandler;
 use EzSystems\PlatformHttpCacheBundle\Handler\ContentTagInterface;
+use EzSystems\PlatformHttpCacheBundle\PurgeClient\PurgeClientInterface;
 
 final class ContentEventsSubscriber extends AbstractSubscriber
 {
+    /** @var \eZ\Publish\SPI\Persistence\Content\Handler */
+    private $contentHandler;
+
+    /** @var bool */
+    private $isTranslationAware;
+
+    public function __construct(
+        PurgeClientInterface $purgeClient,
+        LocationHandler $locationHandler,
+        UrlHandler $urlHandler,
+        ContentHandler $contentHandler,
+        bool $isTranslationAware
+    ) {
+        parent::__construct($purgeClient, $locationHandler, $urlHandler);
+
+        $this->isTranslationAware = $isTranslationAware;
+        $this->contentHandler = $contentHandler;
+    }
+
     public static function getSubscribedEvents(): array
     {
         return [
@@ -93,12 +119,40 @@ final class ContentEventsSubscriber extends AbstractSubscriber
 
     public function onPublishVersionEvent(PublishVersionEvent $event): void
     {
-        $contentId = $event->getContent()->getVersionInfo()->getContentInfo()->id;
+        $content = $event->getContent();
+        $versionInfo = $content->getVersionInfo();
+        $contentType = $content->getContentType();
+        $contentId = $versionInfo->getContentInfo()->id;
 
         $tags = array_merge(
             $this->getContentTags($contentId),
             $this->getContentLocationsTags($contentId)
         );
+
+        $initialLanguageCode = $versionInfo->getInitialLanguage()->languageCode;
+        $mainLanguageCode = $versionInfo->getContentInfo()->mainLanguageCode;
+
+        $isNewTranslation = true;
+        try {
+            $prevVersionInfo = $this->contentHandler->loadVersionInfo($contentId, $event->getVersionInfo()->getContentInfo()->currentVersionNo);
+            $isNewTranslation = !in_array($initialLanguageCode, $prevVersionInfo->languageCodes);
+        } catch (NotFoundException $e) {
+        }
+
+        if (
+            !$this->isTranslationAware ||
+            $isNewTranslation ||
+            ($initialLanguageCode === $mainLanguageCode && !$this->isContentTypeFullyTranslatable($contentType))
+        ) {
+            $this->purgeClient->purge($tags);
+
+            return;
+        }
+
+        $tags = array_map(static function (string $tag) use ($initialLanguageCode): string {
+            return $tag . $initialLanguageCode;
+        }, $tags);
+        $tags[] = ContentTagInterface::CONTENT_ALL_TRANSLATIONS_PREFIX . $contentId;
 
         $this->purgeClient->purge($tags);
     }
@@ -131,5 +185,12 @@ final class ContentEventsSubscriber extends AbstractSubscriber
         $this->purgeClient->purge(
             $this->getContentTags($contentId)
         );
+    }
+
+    private function isContentTypeFullyTranslatable(ContentType $contentType): bool
+    {
+        return !$contentType->getFieldDefinitions()->any(static function (FieldDefinition $fieldDefinition): bool {
+            return !$fieldDefinition->isTranslatable;
+        });
     }
 }
--- a/src/Resources/config/services.yml
+++ b/src/Resources/config/services.yml
@@ -92,6 +92,12 @@ services:
         tags:
             - { name: kernel.event_subscriber, priority: -100 }
 
+    Ibexa\HttpCache\EventSubscriber\AddContentLanguageHeaderSubscriber:
+        arguments:
+            - '%ibexa.http_cache.translation_aware.enabled%'
+        tags:
+            - { name: kernel.event_subscriber }
+
     ezplatform.http_cache.repository_tag_prefix:
         class: EzSystems\PlatformHttpCacheBundle\RepositoryTagPrefix
         # Use config resolver to be able to lazy load reading SA setting "repository" to avoid scope change issues
--- a/.github/workflows/browser-tests.yaml
+++ b/.github/workflows/browser-tests.yaml
@@ -30,6 +30,18 @@ jobs:
             test-setup-phase-1: '--mode=standard --profile=httpCache --suite=setup'
         secrets:
             SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+    varnish-translation-aware:
+        name: "Varnish integration tests (translation-aware)"
+        uses: ibexa/gh-workflows/.github/workflows/browser-tests.yml@main
+        with:
+            project-edition: 'oss'
+            project-version: '^3.3.x-dev'
+            setup: "doc/docker/base-dev.yml:doc/docker/varnish.yml:doc/docker/selenium.yml"
+            test-suite:  '--mode=standard --profile=httpCache --suite=varnish-translation-aware'
+            test-setup-phase-1: '--mode=standard --profile=httpCache --suite=setup-translation-aware'
+            test-setup-phase-2: '--mode=standard --profile=httpCache --suite=setup'
+        secrets:
+            SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
     varnish-token:
         name: "Varnish integration tests with invalidate token"
         uses: ibexa/gh-workflows/.github/workflows/browser-tests.yml@main
--- a/behat_suites.yml
+++ b/behat_suites.yml
@@ -19,7 +19,22 @@ httpCache:
             paths:
                 - '%paths.base%/vendor/ezsystems/ezplatform-http-cache/features/varnish'
             filters:
-                tags: '@varnish' 
+                tags: '@varnish&&~@translationAware' 
+            contexts:
+                - EzSystems\Behat\API\Context\TestContext
+                - EzSystems\Behat\API\Context\ContentTypeContext
+                - EzSystems\Behat\Core\Context\TimeContext
+                - EzSystems\Behat\Core\Context\ConfigurationContext
+                - EzSystems\Behat\API\Context\ContentContext
+                - Ibexa\Behat\Browser\Context\BrowserContext
+                - Ibexa\Behat\Browser\Context\AuthenticationContext
+                - Behat\MinkExtension\Context\MinkContext
+                - Ibexa\Behat\Browser\Context\ContentPreviewContext
+        varnish-translation-aware:
+            paths:
+                - '%paths.base%/vendor/ezsystems/ezplatform-http-cache/features/varnish'
+            filters:
+                tags: '@varnish&&~@translationNotAware' 
             contexts:
                 - EzSystems\Behat\API\Context\TestContext
                 - EzSystems\Behat\API\Context\ContentTypeContext
@@ -38,6 +53,7 @@ httpCache:
                 - EzSystems\Behat\API\Context\ContentTypeContext
                 - EzSystems\Behat\Core\Context\ConfigurationContext
                 - EzSystems\Behat\API\Context\ContentContext
+                - EzSystems\Behat\API\Context\LanguageContext
         setup-token:
             paths:
                 - '%paths.base%/vendor/ezsystems/ezplatform-http-cache/features/setup/invalidateToken.feature'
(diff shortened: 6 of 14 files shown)


ViniTou:
>Unfortunately, I wasn't able to find a way to do that without ugly hacking. Maybe someone has an idea how/if it can be done?

Is this not what you are looking for `\Ibexa\HttpCache\ProxyClient\Varnish`?

@mateuszbieniek

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        if (HttpKernelInterface::MAIN_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+        $content = $request->attributes->get('content');
+        if ($content) {
+            $contentLanguage = $content->prioritizedFieldLanguageCode ?? $content->versionInfo->initialLanguageCode;
```
adamwojs: This is Dependency Inversion Principle violation. `$content->prioritizedFieldLanguageCode` is not part of the PAPI. Moreover, it's marked as internal.

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+    public function onKernelResponse(ResponseEvent $event)
+    {
+        if (HttpKernelInterface::MAIN_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+        $content = $request->attributes->get('content');
+        if ($content) {
```
adamwojs: What guarantee that `$request->attributes->get('content')` is instance of `\eZ\Publish\API\Repository\Values\Content\Content`

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+class AddContentLanguageHeaderListener implements EventSubscriberInterface
```
adamwojs:
```suggestion
final class AddContentLanguageHeaderListener implements EventSubscriberInterface
```

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function onKernelResponse(ResponseEvent $event)
+    {
+        if (HttpKernelInterface::MAIN_REQUEST != $event->getRequestType()) {
```
adamwojs: You should update `symfony/http-kernel` dependency if you like to use `HttpKernelInterface::MAIN_REQUEST`, which is available since Symfony 5.3

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformHttpCacheBundle\EventListener;
```
adamwojs: 1) New classes should be defined under `Ibexa\` namespace

adamwojs: 2) `declare(strict_types=1);` is missing as well as type declarations

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,40 @@
...
+        if (HttpKernelInterface::MAIN_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+        $content = $request->attributes->get('content');
+        if ($content) {
+            $contentLanguage = $content->prioritizedFieldLanguageCode ?? $content->versionInfo->initialLanguageCode;
+            $event->getResponse()->headers->add(['xlang' => $contentLanguage]);
```
adamwojs: Could you please define const with `xlang` literal?

src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
```diff
@@ -93,13 +94,25 @@ public function onHideContentEvent(HideContentEvent $event): void
...
         );
 
+        $initialLanguageCode = $versionInfo->initialLanguageCode;
+        if (
+            $initialLanguageCode !== $versionInfo->getContentInfo()->mainLanguageCode ||
+            $this->isContentTypeFullyTranslatable($contentType)
+        ) {
+            $tags = array_map(function (string $tag) use ($initialLanguageCode) {
+                return $tag . $initialLanguageCode;
+            }, $tags);
```
adamwojs: Nitpick: anonymous function could be marked as static and should declare return type

src/EventSubscriber/CachePurge/ContentEventsSubscriber.php
```diff
@@ -132,4 +145,16 @@ public function onUpdateContentMetadataEvent(UpdateContentMetadataEvent $event):
...
+    private function isContentTypeFullyTranslatable(ContentType $contentType): bool
+    {
+        /** @var \eZ\Publish\API\Repository\Values\ContentType\FieldDefinition $fieldDefinition */
+        foreach ($contentType->getFieldDefinitions() as $fieldDefinition) {
+            if (!$fieldDefinition->isTranslatable) {
+                return false;
+            }
+        }
+
+        return true;
```
adamwojs: Could be replaced with `\eZ\Publish\API\Repository\Values\ContentType\FieldDefinitionCollection::any` call

adamwojs (changes requested):
Does it work with always available flag ?

src/EventListener/AddContentLanguageHeaderListener.php
webhdx: Also `VersionInfo::$initialLanguageCode` is optional and can be `null` in some cases (i.e. when creating content through PAPI).

src/EventListener/AddContentLanguageHeaderListener.php
webhdx: I'm not sure if this should just check `$request->attributes` without any context. Better limit it to content view routes so it's much more predictable.

mateuszbieniek:
> Does it work with always available flag ?

Yes - when content is displayed with AA flag and there is no translation, the main translation is used. In this case, `prioritizedFieldLanguageCode` is set to `null`, and content is tagged using `initialLanguageCode` which will be set to the main translation's language code (I'll change it in regards to @webhdx). Thinking about it, I have to add a case for cache invalidation when a new translation is added - to properly invalidate cache generated with AA flag.

src/EventListener/AddContentLanguageHeaderListener.php
mateuszbieniek: @webhdx I don't think this is necessary - is there any scenario where `content` attribute is set and is an instance of `\eZ\Publish\API\Repository\Values\Content\Content` outside of content view? Also - worst-case scenario - response will have additional tags that won't affect it in any way.

mateuszbieniek:
> > Unfortunately, I wasn't able to find a way to do that without ugly hacking. Maybe someone has an idea how/if it can be done?
> 
> Is this not what you are looking for `\Ibexa\HttpCache\ProxyClient\Varnish`?
> 
> @mateuszbieniek

This looks promising, but I still have to pass `onPublishVersionEvent` context there, so `xlang` is included to only relevant purge requests - which, currently, I have no idea how to do in a non-hackish way.

vidarl (changes requested):
The size limits on headers is a problem. It is not only a problem with limit on the size on one header, but there are also limits on the size of all headers combined. These limits are also different in nginx, apache, varnish and Fastly, ref. https://doc.ibexa.co/en/latest/guide/cache/content_aware_cache/#a-configure-allowing-larger-headers and https://docs.fastly.com/en/guides/resource-limits#request-and-response-limits

With this PR we are at doubling the number of surrogate/x keys, but "to avoid issues with the webserver's maximum header size, they are synthesized inside VCL".
However, we still have limits in varnish and Fastly.
So, if we go ahead with this approach, I think we should:
 - Definitely make it possible to enable/disable using config.
 - Name the tags using language index instead of language code in order to shorten the tags as much as possible: `c36627t2 r36627t2 (...)` instead of `c36627eng-GB r36627eng-GB (...)`
 - Disable by default on 3.3
 - Consider enable by default on 4.x
 - In dev (maybe also prod) mode : check surrogate/x keys headers sizes generated server side and the total size when synthesized inside VCL. Then send flash messages to user if sizes hits some defined limits
 - Make the feature available both on varnish and fastly at the same time. ( or at least add support for it on Fastly within reasonable time)

mateuszbieniek:
Hi @vidarl, thanks for the great input.

Regarding the header size: the webserver header size is not affected, as the original `xkey` header is not modified in any way.
For Varnish, `http_resp_hdr_len` is not affected, as the header is modified after the response. It looks like it might be relevant for `http_resp_size` as my tests show that increased header counts to the overall response size, which might indicate that it is a good idea to increase it from default 32k by a value of max header size of a webserver (8k for apache, 32k for nginx by default).

> In dev (maybe also prod) mode : check surrogate/x keys headers sizes generated server side and the total size when synthesized inside VCL. Then send flash messages to user if sizes hits some defined limits

I don't think it is in this PR scope to do this: as mentioned `xkey` header is not modified for responses and purge requests are already chunked to multiple requests to avoid it.

mateuszbieniek:
@vidarl after an internal talk with the Engineering team we decided to leave language code for `xkey`, as indexes are not reliable and will introduce unnecessary complexity to the code.

Last commits changes:
- `ibexa.http_cache.translation_aware.enabled` parameter (defaults to `false`) for enabling/disabling feature
- `ca` tag which is always purged when content is published. This tag can be used to flag responses which should always be purged when content is modified. For example, pages displaying content in multiple translations at once.
- If the content is `AlwaysAvailable` the main translation will always be purged on publication, to avoid a stale cache when translation is added after AA response is cached.

mateuszbieniek:
Ok, all tests green (CS is a false positive - wants me to change Ibexa back to Ez Systems).
Latest changes:
The cache will be completely purged for content when a new version published is also a new translation.
Re-requesting CR.

alongosz (changes requested):
This branch has now conflicts.

> CS is a false positive - wants me to change Ibexa back to Ez Systems

It is not. CS requirements for the branch use old version of the code style. For now we need to keep it as expected and update entire codebase for at least 2.3 branch and higher to use new name. If we keep CS failing we're risking skipping relevant issue.

.travis.yml
```diff
@@ -34,6 +34,15 @@ matrix:
             - WEB_HOST="http://web"
             - APP_DEBUG=1
             - ENABLE_SYMFONY_CACHE=1
+        - name: "PHP 7.3 Symfony Cache integration tests with translation-aware cache enabled"
```
alongosz: it's worth to point out that Travis is not available for 4.0 and we're moving every CI to GitHub Actions, also for 3.3.

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\HttpCache\EventListener;
```
alongosz: Seems like it rather belongs to EventSubscriber namespace.

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,52 @@
...
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+final class AddContentLanguageHeaderListener implements EventSubscriberInterface
+{
+    public const CONTENT_LANGUAGE_HEADER = 'xlang';
```
alongosz: Shouldn't it be rather `X-Lang` or `x-lang`? It's rather a general convention to separate `X` header with a dash. Seems Ibexa DXP also respects that.

src/EventListener/AddContentLanguageHeaderListener.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        if (!$this->isTranslationAware || HttpKernelInterface::MASTER_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+        $content = $request->attributes->get('content');
+        if ($content && $content instanceof Content) {
+            $event->getResponse()->headers->add([self::CONTENT_LANGUAGE_HEADER => $content->getDefaultLanguageCode()]);
```
alongosz: What concerns me here is association with Content language rather than Site language. Have you tested the case where Content item's default language is different than Site language? In that case what is visible on the front-end SiteAccess would be different from what Content Version serves as its default translation.

src/Resources/config/services.yml
```diff
@@ -92,6 +92,12 @@ services:
         tags:
             - { name: kernel.event_subscriber, priority: -100 }
 
+    Ibexa\HttpCache\EventListener\AddContentLanguageHeaderListener:
+        arguments:
+            - "%ibexa.http_cache.translation_aware.enabled%"
```
alongosz: CS
```suggestion
            - '%ibexa.http_cache.translation_aware.enabled%'
```

alongosz (changes requested):

src/EventListener/AddContentLanguageHeaderListener.php
mateuszbieniek: As it is strongly related to `xkey` header - which is, well, without `-` - I've taken a naming conversion from there. No strong opinion though.

src/EventListener/AddContentLanguageHeaderListener.php
mateuszbieniek: The response is tagged with language code that Content was displayed. For example:
If you have 3 languages: `eng-GB` (main), `pol-PL` and `ger-DE` and SiteAccess configured like this:
```
        site:
            languages: [pol-PL, ger-DE]
```

When Content with only german translation will be accessed in `site` SA, a German translation will be used, as Polish one is not available, and response will be tagged with `ger-DE` language code. Due to that, when the same German translation will be modified, the HTTP cache for the content will be purged. 
To avoid situations, when (using the same example) a German translation is firstly created, then the response is cached and a new Polish translation is created and the cache is not stale, the whole content cache is cleared when the published version is a new translation. This also resolves potential issues with Always Available content.
In addition, when Main Translation is published, and Content is not fully translatable, all other translations will be purged.

src/EventListener/AddContentLanguageHeaderListener.php
adamwojs: @webhdx suggestion is valid. 

> Also - worst-case scenario - response will have additional tags that won't affect it in any way.

In worst-case scenario cache will be purged even if it's not necessary. Sounds like problem you are trying to solve in this PR, isn't it ?

.travis.yml
mnocon: Yep, I'm moving this job right now (https://github.com/ezsystems/ezplatform-http-cache/pull/165 was merged today)

src/EventListener/AddContentLanguageHeaderListener.php
alongosz: > As it is strongly related to `xkey` header - which is, well, without `-` - I've taken a naming conversion from there. No strong opinion though.

Hmm, Maybe xkey has some other special meaning, but:
```
$ curl -i https://127.0.0.1:8000/
HTTP/2 200 
cache-control: max-age=0, must-revalidate, private, s-maxage=86400
content-type: text/html; charset=UTF-8
date: Tue, 18 Jan 2022 12:20:16 GMT
expires: Tue, 18 Jan 2022 12:20:16 GMT
set-cookie: eZSESSID98defd6ee70dfb1dea416cecdf391f58=kd9g2s55f7ddj9h9c56btjd85q; path=/; secure; httponly; samesite=lax
vary: X-User-Context-Hash
vary: X-Editorial-Mode
x-cache-debug: 1
x-cache-tags: ez-all,siso_user_10,siso_menu,c52,l2,siso_pagelayout_footer,content-101,content-102,content-104,content-103,ct42,pl1,p1,p2
x-debug-token: abcdef
x-debug-token-link: https://127.0.0.1:8000/_profiler/abcdef
x-powered-by: Ibexa Commerce v4
x-robots-tag: noindex
```
in this context, it makes `xlang` inconsistent

src/EventListener/AddContentLanguageHeaderListener.php
alongosz: > The response is tagged with language code that Content was displayed. For example: If you have 3 languages: `eng-GB` (main), `pol-PL` and `ger-DE` and SiteAccess configured like this:
> 
> ```
>         site:
>             languages: [pol-PL, ger-DE]
> ```
> 
> When Content with only german translation will be accessed in `site` SA, a German translation will be used, as Polish one is not available, and response will be tagged with `ger-DE` language code. Due to that, when the same German translation will be modified, the HTTP cache for the content will be purged. To avoid situations, when (using the same example) a German translation is firstly created, then the response is cached and a new Polish translation is created and the cache is not stale, the whole content cache is cleared when the published version is a new translation. This also resolves potential issues with Always Available content. In addition, when Main Translation is published, and Content is not fully translatable, all other translations will be purged.

Fair enough, probably prioritizedFieldLanguageCode is pre-populated with information coming from SiteAccessAware layer.
+1

composer.json
```diff
@@ -27,7 +27,7 @@
         "toflar/psr6-symfony-http-cache-store": "^2.2 || ^3.0"
     },
     "require-dev": {
-        "ibexa/ci-scripts": "^0.1@dev",
+        "ibexa/ci-scripts": "dev-test-api-limits as 0.1.x-dev",
```
alongosz: Friendly reminder: this needs to be restored to the previous state before merging

composer.json
```diff
@@ -39,7 +39,8 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\PlatformHttpCacheBundle\\": "src",
-            "EzSystems\\PlatformHttpCacheBundle\\Tests\\": "tests"
+            "EzSystems\\PlatformHttpCacheBundle\\Tests\\": "tests",
+            "Ibexa\\HttpCache\\": "src"
```
alongosz, line 43: On merge up into 4.0 please move it as the first entry. Right now since most of the classes have old ns it wouldn't make sense.

alongosz (approved):
Looks ok, but since you're linking dev branch, is the browser tests failure expected?

mateuszbieniek:
> Looks ok, but since you're linking dev branch, is the browser tests failure expected?

Looks like browser tests are failing for separate reason:
```
  [Composer\Downloader\TransportException]   
  Could not authenticate against github.com  
```

src/EventSubscriber/AddContentLanguageHeaderSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+use eZ\Publish\API\Repository\Values\Content\Content;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+final class AddContentLanguageHeaderSubscriber implements EventSubscriberInterface
+{
+    public const CONTENT_LANGUAGE_HEADER = 'x-lang';
+    private const CONTENT_VIEW_ROUTES = ['_ez_content_view', 'ez_urlalias'];
```
webhdx: Would be nice if this could be extendable since the class is `final` and the const is `private`. There is no possibility to override this setting which means someone would have to duplicate this subscriber to add support for their own routes.

webhdx (approved):
Looks good to me. There is one place where extendibility could be improved.

src/EventSubscriber/AddContentLanguageHeaderSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+        if (!$this->isTranslationAware || HttpKernelInterface::MASTER_REQUEST != $event->getRequestType()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        $content = $request->attributes->get('content');
+        if (
+            $content instanceof Content &&
+            in_array($request->get('_route'), self::CONTENT_VIEW_ROUTES)
+        ) {
```
adamwojs: Can we check if view is instance of `\Ibexa\Core\MVC\Symfony\View\ContentValueView` and `\Ibexa\Core\MVC\Symfony\View\CachableView` ? 

See 
https://github.com/ibexa/http-cache/blob/main/src/lib/EventSubscriber/HttpCacheResponseSubscriber.php#L44-L47 for inspiration.

src/EventSubscriber/AddContentLanguageHeaderSubscriber.php
lserwatka: Side comment: for 4.0.x-dev you will need to rebrand those routes.

adamwojs (approved):

mnocon (approved):
Tested various translation actions with the feature enabled and disabled, the behavious is correct.

QA approved

PR #174, opened 2022-05-05
[CI] Moved to Github Actions

mnocon:
JIRA: https://issues.ibexa.co/browse/IBX-2314

These changes are meant for v2 only and will be reverted when merged upstream (v3.3 and v4 are on Github Actions already).

The REST tests have been moved to Github Actions with as little changes as possible, as v2 is EOM already.


.github/workflows/browser-tests.yaml
```diff
@@ -0,0 +1,17 @@
+name: Browser tests
+
+on:
+    push:
+        branches:
+            - master
```
alongosz: Can be dropped

alongosz (approved):

bogusez (approved):

damianz5 (approved):

PR #175, opened 2022-06-13
IBX-3014: Varnish 7 support

ibexa-yuna:
| Question           | Answer
| **JIRA issue**     | [IBX-3014](https://issues.ibexa.co/browse/IBX-3014)
| **Target version** | 3.3
| **BC breaks**      | no
| **Doc needed**     | yes

**TODO**:
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [ ] Ask for Code Review.


docs/varnish/vcl/varnish7.vcl
```diff
@@ -0,0 +1,335 @@
+// Varnish VCL for:
+// - Varnish 7.1 or higher
```
vidarl, line 2: It would be good if you in the first commit only copied varnsh5.vcl as-is. So that in the this PR I could acutally see changed in the varnish6 vs varnish7...
As it is not, I have to diff them locally in order to see the difference

docs/varnish/vcl/varnish7.vcl
```diff
@@ -0,0 +1,335 @@
...
+    }
+}
+
+sub ez_purge_acl {
+    if (req.http.x-invalidate-token) {
+        if (req.http.x-invalidate-token != req.http.x-backend-invalidate-token) {
+            return (synth(405, "Method not allowed (invalid token)"));
+        }
+    } else if  (!client.ip ~ invalidators) {
+        return (synth(405, "Method not allowed for IP " + client.ip));
```
vidarl: The reason for why the 405 issued was left out on purpose. It can been convenient for developers to see the reason why you get a 405, but maybe not a good reason to give everyone that piece of information. The same applies to the change 3 lines above

Not sure how much of a security problem it would be to reveal it though
@glye  : any comment?

ibexa-yuna: Hey @vidarl 

- This will display only the client's IP i.e. client already knows his own IP (as internal IP should already be transformed via X-Forwarded-For header).
- This is for debugging purposes and will be removed before PR is merged.

docs/varnish/vcl/varnish7.vcl
ibexa-yuna: @vidarl I can force-push this change to branch, however, in PR you will still see a new file added, not a modification of the previous one.

vidarl: yes. but if you first added the file as-is (a carbon copy of varnsish5.vcl, I could see the changes per commit.. The PR overview would still show it as a new file, correct

ibexa-yuna: Rebased

mnocon:
There's a difference in behaviour with the new VCL (on Varnish 7) compared to the old one when a Content Item is in grace.

Setup:
1. User is logged in on the frontend
2. There's a Content Item "TestContentItem" that is cached by Varnish

When the Content Item is edited in the backend the cache is purged. 

When we request the Content Item on the frontend on Varnish 6:
1) First request: old Content Item is served
2) Second request: new Content Item is served, even if we have to wait for it a bit

On Varnish 7 with current VCL:
1) First request: old Content Item is served
2) Second request: old Content Item is served (if the backfround fetch hasn't finished yet)
3) The new Content Item will be displayed only after the background fetch finishes - meaning that user never has to wait to see Content, but sees the stale content longer.

I believe the new VCL doesn't have logic correspodning to this part: https://github.com/ezsystems/ezplatform-http-cache/blob/2.3/docs/varnish/vcl/varnish5.vcl#L91-L94 in the old VCL.
Meaning that Content Items in grace are served for authenticated users.

I thought about this a bit and I think that it's ok - if we were 100% serious about not letting logged in users see the old Content on the frontend then we wouldn't use soft-purging. It does not make a difference for Admin panel, because HTTP cache is disabled there.

If you agree then I will see if we can adapt the failing test for Varnish 7 (or maybe disable it if it's not possible)

vidarl:
Hi

>When we request the Content Item on the frontend on Varnish 6:
>1.   First request: old Content Item is served
>2.    Second request: new Content Item is served, even if we have to wait for it a bit

I am surprised by "1." The code you refer to should disable the background fetch. Also, when I test myself, I see the behavior I expect, not the one you are observing.

I agree with you. I think the new behaviour is anyway acceptable. I also had a look at the Fastly implementation and it does not have the varnish6 behavior where stale-cache is not served users with cookie.

Implementing this behavior in varnish 7 would complicate things quite a bit, it would require a new restart as `return (miss);` is no longer allowed from `vcl_hit()`. This will add additional complexity which we should be avoid unless it is strictly  speaking needed.

mnocon (approved):
I've tested the behaviour of the application when Varnish 7 is used and haven't found anything odd, so once this gets reviewed it's +1 from me 💪

docs/varnish/vcl/parameters.vcl
```diff
@@ -5,7 +5,7 @@
 // https://github.com/ezsystems/ezplatform/blob/master/doc/docker/entrypoint/varnish/parameters.vcl
 
 backend ezplatform {
-    .host = "127.0.0.1";
+    .host = "web"; // Replace with hostname/ip of the application server
```
vidarl: I don't think  these changes are relevant. They look docker specific. So I would recommend leaving them as-is

The paramters.yml file used by docker containers are anyway this one  https://github.com/ezsystems/ezplatform/blob/master/doc/docker/entrypoint/varnish/parameters.vcl

vidarl (changes requested):
Hi

The vcl changes looks good to me.
But changes in parameters.vcl should be reverted IMO

While testing it, I found however this problem : https://issues.ibexa.co/browse/IBX-3715
But that is not related this PR directly ( both varnish6 and fastly is affected by the same problem )

lserwatka:
@vidarl thank you for your review. @ibexa-yuna we must look into it as soon as you are back. This is top priority.

docs/varnish/vcl/parameters.vcl
ibexa-yuna: Hi Vidar,

As for those changes, they are required to make tests pass properly.
Specifically in https://github.com/ibexa/docker/pull/7/files

However, we should simply refresh that PR to use the file mentioned above, and we should be good to go.

PR #180, opened 2024-04-22
Improvements to Varnish configuration

audunmg:
Hi,
I'm happy to see you're using Varnish, but it seems the VCL configuration is a bit outdated.

I've made a few commits I hope will be helpful.

Summary:
* Varnish versions 5 and older are really not recommended for production anymore
* Grace/keep is automatically handled in Varnish 6, and when backends are healthy, I suggest to cap grace to serve fresh content.
* The official way to detect hit/miss in varnish, more info:
  * https://docs.varnish-software.com/tutorials/hit-miss-logging/
  * https://github.com/varnish/toolbox/tree/master/vcls/hit-miss)
  * https://info.varnish-software.com/blog/using-obj-hits
* using vmod cookie for cookie handling is much more readable than regex, and more performant.

Lastly, I've added some simple varnishtest testcases, these can be ran with the varnishtest utility which comes in the same package as varnish, but it's most powerful when ran using the official docker image in CI.
The tests I've added are just one for checking that the configuration is correct, compiles and passes one requet, and one  for demonstrating cookie handling, but more advanced tests can be made, for example to test purging and caching of specific requests and responses.

Please let me know what you think, and if you have inputs to this I'm happy to make changes to make it better for you.

**TODO**:
- [ ] Ask for Code Review.


tests/varnishtests/001-simple-request.vtc
```diff
@@ -0,0 +1,20 @@
...
+shell {
+    sed "s/.host =.*/.host = \"${s1_addr}\";/;s/.port = .*/.port = \"${s1_port}\";/" ${pwd}/docs/varnish/vcl/parameters.vcl > ${tmpdir}/parameters.vcl
+    echo "${s1_addr} ${s1_port}"
+} -run
+
+varnish v1 -arg "-p vcl_path=${tmpdir} -f $PWD/docs/varnish/vcl/${vclfile}" -start
+
+client c1 {
+    txreq
+    rxresp
```
gquintard, line 19: can we get an `expect` here, for the status, or maybe for the `x-cache` header?

